PR #326, opened 2023-08-06
Allow tags for challenges

jhechtf:
Trying to figure out the best way to implement the tags and which tags to allow.


jhechtf:
For right now this seems a little out of scope so I'm going to close this PR. Will keep the branch up in case we need the work again though.

PR #344, opened 2023-08-13
fix: feature card render on firefox

mirashif:
Removed feature card blur filter to fix render on firefox


mirashif:
This was fixed in #373 🎉

PR #380, opened 2023-08-22
Moving monaco to a monorepo

ggrandi:
~~Current~~ Past issues:

~~1. for whatever reason, nextjs doesn't recognize the package so I keep getting `Module not found: Can't resolve '@repo/monaco'` errors~~
~~1. for whatever reason, `pnpm typecheck` has completely stopped working in the `apps/web` directory~~
~~1. the `useLocalStorage` hook is needed both in the monaco package and not and I don't know where to put it~~

Should all be fixed now

the big diff is vendoring the dependencies


Hacksore (approved):
Approving but should this be failing?
https://vercel.com/trash-company/typehero-admin/9sPdFrU8rw2a4rzdcjaodKQCKtXU

ggrandi:
i genuinely cannot figure out why the admin stuff is broken

apps/admin/src/app/challenge/[challengeId]/challenge-review.tsx
```diff
@@ -70,7 +68,7 @@ export function ChallengeReview({ challenge }: Props) {
           <Markdown>{challenge.description}</Markdown>
         </div>
         <div className="flex-1 rounded-l-xl rounded-r-2xl border border-zinc-300 dark:border-zinc-700">
-          <CodeEditor value={challenge.tests} />
+          {/* <CodeEditor value={challenge.tests} />   */}
```
Comment (Hacksore) [inline apps/admin/src/app/challenge/[challengeId]/challenge-review.tsx:71]: @ggrandi do we know why this is broken? I believe trash needs to use it so maybe we should make sure this works before merge.

bautistaaa (approved):

PR #383, opened 2023-08-22
fix: contributing commands

QuiiBz:
- we have to run both `db:push` and `db:seed`, and not the `seed` task that doesn't exists
- we have to use `pnpm dev` and not `turbo dev` because of the `--concurrency` option inside the root `package.json`


jhechtf (approved):
LGTM

Issue #243, opened 2023-07-16
[a11y] focus: and focus-visible: states


PickleNik:
- add focus:outline-none on default html elements
 - use focus-visible:ring-2 instead (or change border if component already has a border set)
 - match hover effects wherever it makes sense
 - set ring-blue-600 or something to match primary color
 - override all shadcn input fields to not have focus-visible:ring-offset-x

laureanray:
Can I work on this @PickleNik @bautistaaa?

Milind-Jamnekar:
If it is solved why not close this issue @bautistaaa ? and there are 1 or more issue like this (merged but not closed)

laureanray:
As I've mentioned in the PR that was closed A11y is like a moving target as we add more components/features. The PR addressed some but not all the a11y stuff.

laureanray:
I no longer have the time to work on this, someone else can pick it up. Apologies!

PickleNik:
no worries, huge thanks for all your work so far!

> I no longer have the time to work on this, someone else can pick it up. Apologies!

MichaelG824:
@PickleNik @bautistaaa Could I pick this up?

PR #388, opened 2023-08-22
Feat/a11y focus and focus visible states

laureanray:
I think a11y is a moving target as we create more features/components. I will open the PR for now and maybe just do a follow up if I found new elements to add the proper classes. 

We should probably keep #243 open for tracking. 

# Preview:
Note: I have used "force element state" to apply the `:focus` and `:focus-visible` pseudoclasses. 

On focus, the elements should look like the screenshot (obviously not at the same time) 

## Navigation
image

## Features
image

## Community
image

## Explore
image
image

## Other 
- This should also fix some elements being focused twice because of nested focus-able elements like `<Link>` and `<Button>` nested.
- Removed shadcn components `focus-visible:ring-offset-2`


PickleNik:
@laureanray I fixed a couple conflicts I introduced in main, pls pull, sorry, thanks

PickleNik:
This is huge, thanks @laureanray.

I wanna run it thru my local to checkout first but then I'll approb and merge soon as I can 👍

PickleNik:
created a separate issue for some related but out-of-scope stuff if u wanna take it #502

laureanray:
> created a separate issue for some related but out-of-scope stuff if u wanna take it #502

Yes, you can assign it to me.

PR #390, opened 2023-08-22
Prevented unnecessary query

typesafegh:
Prevented unnecessary query when accessing typehero.dev/[username] without having the @ in the username

PR #391, opened 2023-08-23
docs: correct some of the steps in the contributing guide

dylf:
Minor corrections to the contributing guide

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,8 +73,8 @@ pnpm install
 ### 6. Push Database Schema and Seed
 
 ```
-turbo db:push
-turbo db:seed
+pnpm db:push
+pnpm db:seed
 ```
 
 ### Running the dev server


Hacksore (commented):
Thanks for the contribution!

I don't believe we should install `turbo` globally because it's better for us to depend on the locally installed version. 

Would be helpful to make sure we are all running the same versions as bugs are bound to pop up more and more with the rest rewrite.

https://turbo.build/repo/docs/installing#install-per-repository

Also the env vars were fixed via #390

dylf:
@Hacksore good call. I agree it would just be best to use the project version of turbo.

I tweaked this to just update the guide to use the pnpm scripts for the db commands, but feel free to close it if you add that change elsewhere.

Hacksore (approved):

Hacksore (approved):

dylf:
accidentally snuck in a formatting change there :upside_down_face:

PR #405, opened 2023-08-24
Add feature flags

Hacksore:
this add feature flags and only allows you to to `/` and `/waitlist` hypothetically.

The edge DB has this for all envs set currently:
```
{
  "exploreButton": false,
  "loginButton": false
}
```

### DIFF
--- a/apps/web/src/app/api/flags/route.ts
+++ b/apps/web/src/app/api/flags/route.ts
@@ -1,0 +1,7 @@
+import { NextResponse } from 'next/server';
+import { getAllFlags } from '~/utils/feature-flags';
+
+export async function GET() {
+  const flags = await getAllFlags();
+  return NextResponse.json(flags);
+}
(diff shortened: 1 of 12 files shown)


apps/web/src/app/api/flags/route.ts
```diff
@@ -0,0 +1,7 @@
+import { NextResponse } from 'next/server';
+import { getAllFlags } from '~/utils/feature-flags';
+
+export async function GET() {
```
jhechtf, line 4: should fetching of the feature flags maybe be done in the middleware layer?

Hacksore: It could be done there but also really only tryna keep you from accessing anything other than `/` and `/waitlist`.

So we could move all the logic to the middleware for evaluation but I'll have to play around with that.

jhechtf: Fair enough.

jhechtf (approved):
LGTM re @bautistaaa

PR #408, opened 2023-08-24
Checks adjustments

anthonyshew:
- Removes precommit.
- Adjust linting and formatting commands.


PickleNik (approved):
GLTM

apps/admin/package.json
```diff
@@ -7,8 +7,10 @@
     "build": "next build",
     "clean": "rm -rf .next dist node_modules",
     "dev": "dotenv -e ../../.env next dev -- -p 3001",
-    "format": "prettier -w \"src/**/*.{ts,tsx,css,cjs,mjs,json}\"",
-    "lint": "eslint .",
+    "format": "prettier . --check --cache --cache-location='node_modules/.cache/prettiercache'",
```
jhechtf, line 10: IDK if this is just me, but this creates a new directory in a bunch of places that is named `'node_modules/.cache/prettiercache'`. I believe the intent here is to have the cache stored in the `node_modules/.cache/prettiercache`, which might make more sense to change this to 

```suggestion
"format": "prettier . --check --cache --cache-location=\"node_modules/.cache/prettiercache\"",
```

anthonyshew: Are you on Windows? I wasn't aware of that, will change.

.eslintrc.js
```diff
@@ -2,6 +2,7 @@
 module.exports = {
   root: true,
   extends: [require.resolve('./tooling/config-eslint/node')],
+  ignorePatterns: ['apps/**', 'packages/**', 'tooling/**'],
```
jhechtf, line 5: Correct me if I'm wrong but doesn't this remove linting from the majority of the repo?

anthonyshew: This file only gets used by the root of the repository. In workspaces, you'll find `.eslintrc.js` files that have `root: true` so they won't resolve any rules higher than there.

I do seem to have a misconfiguration elsewhere, though. 😄

anthonyshew: Ah, nope, I'm all good. Forgot that we have a lot of rules turned off right now in `tooling/config-eslint/utils/rules.js`.

If you want to see the linter get more strict, you can comment out some rules in there and run `turbo lint --force` and you'll be given issues.

jhechtf (approved):
LGTM

Hacksore:
> GLTM

GOOD LOOKING TO ME

PR #411, opened 2023-08-24
Test turbo ignore

Hacksore:
Was trying to skip builds of apps that have not changed. I edited the git settings for "trash company" but it rebuild/deployed everything 😅.

image


anthonyshew:
It will rebuild everything on a new branch. Have you tried redeploying with a small change after to see if it is cancelling?

Hacksore:
@anthonyshew did you delete your comment 👀

Hacksore:
> It will rebuild everything on a new branch. Have you tried redeploying with a small change in an app to see if it cancels the others?

See latest commit it only skipped one

QuiiBz:
> See latest commit it only skipped one

It makes sense because both `apps/web` and `apps/admin` depend on the `@repo/emails` package that has been updated in this PR. Also `apps/emails` should be moved to `packages/emails`, right? It's not really an application since it is imported by other apps.

Hacksore:
@QuiiBz we actually are using it as an app to host static assets 😅.

if we want to move it to packages we need update where the image is loaded from. Probably just use typehero.dev to load the static assest when deployed.

However the main reason I had it like this was so you can play around with local assets but we could make it load from `http://localhost:3000`.

QuiiBz:
Yeah, that's probably a good option, having apps that depend on other apps seems wrong.

Hacksore:
hopefully this is fixed via #445

PR #422, opened 2023-08-25
[POC] feat: GitLab auth PoC

jhechtf:
## Description

Added in some rough testing for GitLab OAuth flow.

## Notes

The Gitlab user comes back with a `created_at` which is an integer, so I had to add that.

Whether or not we want to do that we can.

Some other limitations: you cannot have more than 1 oauth source that share the same email, and from what I've read in next auth there is no work around because that was a decision by the next auth team.

### DIFF
--- a/apps/web/src/components/ui/navigation.tsx
+++ b/apps/web/src/components/ui/navigation.tsx
@@ -23,6 +23,8 @@ export function Navigation() {
   const pathname = usePathname();
   const featureFlags = useContext(FeatureFlagContext);
 
+  console.info(featureFlags);
+
   return (
     <header className="z-10 w-full">
       <nav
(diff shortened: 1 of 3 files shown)


Hacksore:
That last point has bitten me so many times - surely there has to be a workaround to allow me@domain.com to login via GitHub/<insertSeoncdayProvider>.

imo I’d think discord login would be much more valuable as almost everyone has it.

jhechtf:
> That last point has bitten me so many times - surely there has to be a workaround to allow [me@domain.com](mailto:me@domain.com) to login via GitHub/`<insertSeoncdayProvider>`
> 
> imo I’d think discord login would be much more valuable as almost everyone has it.

Oh hell yeah hadn't even thought about Discord. Hol' up

Hacksore:
Pretty sure the biggest blocker is figuring out the best UX for supporting multiple providers.

like my GitHub and discord are the same email and I’ll error out 💀.

jhechtf:
Closing as @bautistaaa doesn't love GitLab or Discord :p

PR #425, opened 2023-08-25
feat: integrate million js automode

kunal00000:
Resolves https://github.com/aidenybai/million/issues/556
/claim https://github.com/aidenybai/million/issues/556


jhechtf:
I pulled down these changes and it breaks local dev -- were you able to get it to work on your local?

kunal00000:
Yes it is breaking when I checked it now, let me try to fix it.

PR #426, opened 2023-08-25
use mock feature flags locally

bautistaaa:
band aid for now


apps/web/src/app/feature-flag-provider.tsx
```diff
@@ -6,10 +6,23 @@ export const FeatureFlagContext = createContext<Record<string, boolean>>({});
 interface Props {
   children: ReactNode;
 }
+
+const mockFlags = {
```
QuiiBz, line 10: We'll probably need a proper type instead of using `Record<string, boolean>` in the context and here

Issue #399, opened 2023-08-23
Add some vim keybinds


ggrandi:
Keybinds I can think of (might add more):
- [x]  `K`: hover functionality (#403)
- [x] (tbd): code diagnostics (#434)
- [x] `gd`: goto definition (#434)
- [x] `gr`: goto references (#434)
- [x] `gi`: goto implementation (#434)

QuiiBz:
We might want this one too:
- `gi`: goto implementation

Hacksore:
# can we please have `gcc` added to block comment as a GFI?

@bautistaaa @ggrandi

PR #434, opened 2023-08-26
feat: add more vim keybindings to editor

dylf:
Added more vim keybinds related to #399 

Go to references is a bit odd because it opens a preview window that makes you put your hand back on the mouse :sweat_smile:

### DIFF
--- a/packages/monaco/src/vim-keybindings.ts
+++ b/packages/monaco/src/vim-keybindings.ts
@@ -1,0 +1,54 @@
+import { VimMode } from 'monaco-vim';
+
+type Keybind = {
+  key: string;
+  action: string;
+  name: string;
+};
+
+function registerKeybind({ key, action, name }: Keybind) {
+  VimMode.Vim.defineAction(name, function (ctx) {
+    ctx.editor.trigger('action', action, null);
+  });
+  VimMode.Vim.mapCommand(key, 'action', name);
+}
+
+const KEYBINDS: Keybind[] = [
+  {
+    key: 'K',
+    action: 'editor.action.showHover',
+    name: 'hover',
+  },
+  {
+    key: 'gd',
+    action: 'editor.action.goToDeclaration',
+    name: 'goToDeclaration',
+  },
+  {
+    key: 'gr',
+    action: 'editor.action.goToReferences',
+    name: 'goToReferences',
+  },
+  {
+    key: 'gi',
+    action: 'editor.action.goToImplementation',
+    name: 'goToImplementation',
+  },
+  {
+    key: 'gx',
+    action: 'editor.action.openLink',
+    name: 'goToLink',
+  },
+  {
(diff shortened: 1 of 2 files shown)


ggrandi:
I'm super impressed! how did you find out how to use the `editor.trigger` function?

packages/monaco/src/vim-keybindings.ts
```diff
@@ -0,0 +1,38 @@
+import { type VimMode } from 'monaco-vim';
+
+export default function registerKeyBindings(vim: typeof VimMode) {
```
QuiiBz, line 3: I think we prefer named exports instead of default exports

packages/monaco/src/vim-keybindings.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  vim.Vim.defineAction('nextDiagnostic', function (ctx) {
+    ctx.editor.trigger('action', 'editor.action.marker.next', null);
+  });
+  vim.Vim.mapCommand(']d', 'action', 'nextDiagnostic');
+
+  vim.Vim.defineAction('prevDiagnostic', function (ctx) {
+    ctx.editor.trigger('action', 'editor.action.marker.prev', null);
+  });
+  vim.Vim.mapCommand('[d', 'action', 'prevDiagnostic');
```
QuiiBz, line 37: We can probably avoid duplicating the same things over and over by adding a utility function and looping over a list keybinds. Something like:

```suggestion
  type Keybind = {
    keybind: string
    name: string
    action: string
  }

  function registerKeyBinding({ keybind, name, action }: Keybind) {
    vim.Vim.defineAction(name, ctx => {
      ctx.editor.trigger('action', action, null);
    });
    vim.Vim.mapCommand(keybind, 'action', name);
  }
  
  const KEYBINDS: Keybind[] = [...];
  KEYBINDS.forEach(registerKeyBinding);
```

Hacksore:
> I'm super impressed! how did you find out how to use the `editor.trigger` function?

prolly a combination of game knowlege and TS is my guess.
image

dylf:
> I'm super impressed! how did you find out how to use the `editor.trigger` function?

Lol banging my head against some of the monaco stuff for a bit. I couldn't get the lsp actions working with `getActions('..').run()` but found trigger worked there and without that setTimeout

packages/monaco/src/vim-keybindings.ts
dylf: Me too, updating as per your pr notes

dylf:
@QuiiBz Updated with your suggested feedback. It turns out I didn't really need to export anything at all I just pulled the `VimMode` class directly into the keybinds file.

ggrandi (approved):

QuiiBz (approved):

Issue #433, opened 2023-08-26
[Bug]: Selecting challenge in the learning tracks card (Landing Page) causes gradient to solidify


basokant:
### What happened?

https://twitter.com/iamyeizi/status/1695241048195273170?s=20

![image](https://github.com/bautistaaa/typehero/assets/70033855/4da83e0b-2571-44fa-8f91-1afef815d9f3)

Has something to do with the absolutely positioned backdrop blur element that is used for the gradient effect on each challenge item.

Seems to be Safari only.

### What browsers are you seeing the problem on?

Safari

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #437, opened 2023-08-26
Solve - Selecting challenge in the learning tracks card (Landing Page) causes gradient to solidify

maxdemaio:
For #433

This removes the blurred background gradient on the absolutely positioned difficultly `div` for the right text. Then, it adds a background gradient from right to left with an opacity of the difficulty's background color on the outer `div`. I checked this works in light/dark mode on safari. Would be glad to make any needed changes!

Light mode:
image

Dark mode:
image

### DIFF
--- a/apps/web/src/components/landing/feature-card.tsx
+++ b/apps/web/src/components/landing/feature-card.tsx
@@ -538,9 +538,6 @@ function Track({ className, difficulty, id, label }: Track) {
         <div
           className={`relative text-xs font-medium tracking-wide ${COLORS_BY_DIFFICULTY[difficulty]}`}
         >
-          <div
-            className={`absolute right-0 top-1/2 h-12 w-12 -translate-y-1/2`}
-          />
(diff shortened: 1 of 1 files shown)


apps/web/src/components/landing/feature-card.tsx
```diff
@@ -539,7 +539,7 @@ function Track({ className, difficulty, id, label }: Track) {
           className={`relative text-xs font-medium tracking-wide ${COLORS_BY_DIFFICULTY[difficulty]}`}
         >
           <div
```
PickleNik, line 541: delet dis element

PickleNik (changes requested):

apps/web/src/components/landing/feature-card.tsx
maxdemaio: You got it!

PickleNik (approved):
LGTM

Issue #441, opened 2023-08-26
Light mode Waitlist Banner looks goofy, pls fix


PickleNik:
![image](https://github.com/bautistaaa/typehero/assets/31113245/b8826f61-415e-4e2a-ab14-153dbfd8d1f0)

Mahmoudgalalz:
Taking it

maxdemaio:
Addressed this in https://github.com/bautistaaa/typehero/pull/443 👍🏻

PR #442, opened 2023-08-26
Changed waitlist font color

Dilandds:
#441 text color changed to look better.


QuiiBz (changes requested):
I think you updated the wrong line? It should be here: https://github.com/bautistaaa/typehero/pull/442/files#diff-72aab63156427c6dcdfb26818ab4b944a195af51e0cf645e7898a71b83e28ff0R182

bautistaaa:
merged another that had a similar fix. 

you can just say text-white vs explicitly defining dark. thanks for the PR!

PR #446, opened 2023-08-26
added prettier to the editor

ggrandi:
so far I have a added a sensible prettier config and bumped the prettier version

next steps:
- Add the ability to change the prettier config (gfi)


packages/monaco/src/prettier.ts
```diff
@@ -0,0 +1,34 @@
...
+import estree from 'prettier/plugins/estree';
+import typescript from 'prettier/plugins/typescript';
+import type { Options } from 'prettier';
+import type * as monaco from 'monaco-editor';
+
+// TODO: use the settings store for the config
+const prettierConfig = {
+  plugins: [estree, typescript],
+  parser: 'typescript',
+  useTabs: true,
```
Hacksore, line 11: controversial setting 😬 

Maybe you could make this use the current editor setting?

Hacksore (approved):

packages/monaco/src/prettier.ts
ggrandi: I think we already control the tab size with the `settings-store` so it shouldn't be an issue

Plus this saves us many bytes of space in the db

Hacksore: oh we don't even support spaces lmao
image

ggrandi: I think we support whatever the editor uses which might default to spaces

PR #447, opened 2023-08-26
Floating Navbar

manuarora700:
- Floating fixed navbar with a backdrop
- Navbar disappears if the user scrolls down
- Navbar comes back up if the user scrolls up


apps/web/src/components/ui/navigation.tsx
```diff
@@ -23,10 +24,42 @@ export function Navigation() {
...
+    [0, 0.75, 1],
+    [0, 0, 1],
+  );
+
   return (
-    <header className="z-10 w-full">
+    <motion.header
+      className="fixed inset-x-0 z-20 flex w-full overflow-hidden bg-white/50 backdrop-blur-md dark:bg-black/50"
+      style={{
+        height: useTransform(scrollYBoundedProgressThrottled, [0, 1], [80, 0]),
```
PickleNik, line 58: changing height of the element is not ideal for fps as it triggers a bunch of repaints, I'd `translate-y` it off the screen instead

https://github.com/bautistaaa/typehero/assets/31113245/a8631154-6d79-4258-a7d0-3d770c7036e4

PickleNik (changes requested):

PickleNik:
overall I like the idea, what if it also was pill-shaped like on https://diagram.com/?

https://github.com/bautistaaa/typehero/assets/31113245/ad2cdad7-69c2-4e0f-baa2-51d08fc15f1b

apps/web/src/components/ui/navigation.tsx
PickleNik: would look even nicer if it's like `scale(0.9)` and `opacity: 0` alongside `translate-y`, none of those trigger repaints either so it shouldn't impact performance

PickleNik:
sorry i clicked update branch that rebased to main, should not be a big deal, if u got changes just force push and don't worry about it, if u don't just pull before changes, thank you, sorry

bautistaaa:
is this still being worked on?

manuarora700:
yes -- will push an update today @bautistaaa

PickleNik:
> overall I like the idea, what if it also was pill-shaped like on https://diagram.com/?
> 
>  Screen.Recording.2023-08-26.at.2.06.55.PM.mov

this was just an idea btw, it doesn't have to be in this PR, I'd mainly just fix the bugs with it first and then iterate on design later

bautistaaa:
this PR seems dead. going to close

PR #452, opened 2023-08-26
Add Comment Upvotes

### LINKED ISSUE #431: [Comments]: Ability to Upvote a comment
### Description

On the comment component (in the challenge and solution tabs within a challenge page), a user should be able to upvote a comment. The number of upvotes should display within a comment.

### Explanation

We should follow a similar style to Leetcode comment upvoting UI, except we will only allow upvoting:

image

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

basokant:
closes #431 

* adds comment ids to seed data
* generic vote component that is now used for challenges and comments
* change vote to use optimistic updates

image

demo: https://drive.google.com/file/d/1WAn_MlpNg0_AZa9XWTWK64NH5HY1l_Lf/view?usp=sharing


apps/web/src/app/challenge/increment.action.ts
```diff
@@ -0,0 +1,43 @@
...
+        challengeId: rootType === 'CHALLENGE' ? id : undefined,
+        commentId: rootType === 'COMMENT' ? id : undefined,
+        sharedSolutionId: rootType === 'SHAREDSOLUTION' ? id : undefined,
+        submissionId: rootType === 'SUBMISSION' ? id : undefined,
+        userId: session?.user.id,
+      },
+    });
+  }
+
+  if (session?.user && voteExists && !shouldIncrement) {
```
jhechtf, line 36: Just to verify for myself since i haven't looked into this too hard: we only track upvotes, e.g. we don't have the concept of "downvote" as it exists on like reddit / stack overflow, which is why when we are "downvoting" something we just delete the vote?

basokant: yeah, that's correct (you wrote that lolll, just moved the file since we'll be using it for a lot of things in challenges.)

packages/db/mocks/comment.mock.ts
```diff
@@ -5,9 +5,14 @@ import { trashId } from '../seed';
...
+export default function CommentMock(
+  commentNumber: number,
+  parentId?: number,
+): Prisma.CommentCreateManyInput {
   return {
-    text: 'here is a comment',
+    text: `here is a comment ${
+      !parentId ? commentNumber : `reply ${commentNumber} to parent ${parentId}`
+    }`,
     userId: trashId,
```
jhechtf, line 16: for a random user i think in the seed script we pull all the user IDs, so you could use the `faker.helpers.arrayElement` to choose a random userId instead of hard-coding it to trash's

packages/db/mocks/vote.mock.ts
```diff
@@ -1,8 +1,10 @@
-import { type Prisma } from '@prisma/client';
 import { faker } from '@faker-js/faker';
+import { type Prisma } from '@prisma/client';
 
 export function VoteMock(userIds: string[]): Prisma.VoteCreateManyInput {
   return {
     userId: faker.helpers.arrayElement(userIds),
+    rootType: 'CHALLENGE',
+    challengeId: 2,
```
jhechtf, line 8: Same thing as the comment.mock.ts i made -- you can definitely rig up the seed script to pull a random challenge ID so that challenge 2 is the only one with things.

packages/email/package-lock.json
```diff
@@ -0,0 +1,7560 @@
+{
```
jhechtf, line 1: we probably want to get rid of this file since we don't use NPM

jhechtf (changes requested):
delete the random lock file and i everything LGTM

jhechtf (approved):
LGTM

apps/web/src/app/challenge/_components/comments/index.tsx
```diff
@@ -23,6 +24,7 @@ export function Comments({ rootId, type }: Props) {
   const [text, setText] = useState('');
   const commentContainerRef = useRef<HTMLDivElement>(null);
   const queryClient = useQueryClient();
+  const session = useSession();
```
bautistaaa, line 27: im on my phone but is this being used

apps/web/src/app/challenge/increment.action.ts
```diff
@@ -0,0 +1,43 @@
+'use server';
+import { getServerAuthSession } from '@repo/auth/server';
+import { prisma } from '@repo/db';
+import type { VoteType } from '@repo/db/types';
+
+export async function incrementOrDecrementUpvote(
+  id: number,
+  rootType: VoteType,
+  shouldIncrement: boolean,
```
bautistaaa, line 9: i know i did the original version but do we need this boolean. the existence of the record should imply to delete it right? cant remember if i had this here cause the user can spam clicks

basokant: o wait u wrote this lolll I thought @jhechtf did. yeah I think that was what ur intention was, which is why I didn't touch it.

bautistaaa: ok maybe you wrote it 🤣

but i created the initial version at least

apps/web/src/app/challenge/_components/comments/index.tsx
basokant: yeah it isn't used, removed it

apps/web/src/app/challenge/increment.action.ts
bautistaaa: does it still work without that boolean

basokant: I think it should because we're checking if it exists or not

packages/db/mocks/vote.mock.ts
basokant: hmmm ok, I'll do this in a separate PR

packages/db/mocks/comment.mock.ts
basokant: yeah I'll do this in a separate PR lol

packages/db/mocks/vote.mock.ts
bautistaaa: i hard coded 2 so i could always just point to one challenge to test things 😀

Hacksore:
> delete the random lock file and i everything LGTM

FYI the random lock file thing is related to https://github.com/resendlabs/react-email/pull/892

PR #456, opened 2023-08-27
feat: use `next/image` for contributors avatars

QuiiBz:
We were using a `div` with a `backgroundImage`, which causes big payloads and unnecessary network requests when not in the viewport:

Screenshot 2023-08-27 at 17 39 44


bautistaaa (approved):
ohhhh

laureanray (approved):

PR #464, opened 2023-08-27
Some cleanup on commands.

anthonyshew:


Hacksore (approved):
:shipit:

packages/monaco/src/vim-mode.tsx
```diff
@@ -2,7 +2,8 @@
 
 import type * as monaco from 'monaco-editor';
 import { useEffect, useRef } from 'react';
-import { VimMode, initVimMode } from 'monaco-vim';
+import type { VimMode } from 'monaco-vim';
```
bautistaaa, line 5: why separate

Hacksore: type

PR #468, opened 2023-08-28
Fix: Comment pagination is broken

maxdemaio:
This PR addresses https://github.com/bautistaaa/typehero/issues/453

Previously, the `totalReplies` were calculated from the comments. But, since the comments are paginated, the `totalReplies` was 20. This is because there were 10 items per page in that instance, and then 2 replies on each comment.

But, we want the `totalReplies` across all pages (100) and add that amount to the `totalComments` (50) which would be 150. To achieve this, I added a separate count query to calculate the total replies where the parent ID isn't null. Updated:

image

(shoutout manpreet and kampouse)


maxdemaio:
For the prettier check, looks like it's because:

```
A dynamic import callback was not specified.
TypeError: A dynamic import callback was not specified.
```

However, this PR does not change any imports 👍🏻

apps/web/src/app/challenge/_components/comments/getCommentRouteData.ts
```diff
@@ -28,6 +29,17 @@ export async function getPaginatedComments({
...
+    where: {
+      rootType,
+      parentId: {
+        not: null,
+      },
+      visible: true,
+      ...(rootType === 'CHALLENGE' ? { rootChallengeId: rootId } : { rootSolutionId: rootId }),
+    },
+  });
+
```
basokant, line 42: wish we could do this and totalComments in one query.... anyway this looks good!

apps/web/src/app/challenge/_components/comments/getCommentRouteData.ts
```diff
@@ -71,8 +83,6 @@ export async function getPaginatedComments({
 
   const totalPages = Math.ceil(totalComments / PAGESIZE);
 
-  const totalReplies = comments.reduce((a, c) => a + c._count.replies, 0);
-
   return {
     totalComments: totalReplies + totalComments,
```
basokant, line 87: I'm hoping this doesn't break pagination, trust that you tested it but if it's broken we can always fix it

basokant (approved):
LGTM

apps/web/src/app/challenge/_components/comments/getCommentRouteData.ts
maxdemaio: Agreed! If we ever find a way I'd be glad to refactor 👍🏻

basokant (approved):
lgtm (again)

PR #469, opened 2023-08-28
Comment Sorting

### LINKED ISSUE #341: [Comments] Ability to sort
sort by recent, upvotes, number of replies(?)

basokant:
closes #341 

Add comment sorting functionality and UI to the comments component. This adds a select dropdown for the sortKey (createdAt, votes, or replies) and the sortOrder (asc, desc).

image

demo: https://drive.google.com/file/d/1a9tS4J2ShwiJQ5tvcDTEj8Pu9AoB9eW5/view?usp=sharing


jhechtf (approved):
Codewise LGTM, only nit is that I would probably prefer the ascending / descending thing by a single button that toggles between the two but it's not something i feel strongly about so we can ship it how it is and fix it later if it ends up annoying us enough

PickleNik (approved):
LFGTM

PR #489, opened 2023-08-28
[#478] feat: reset button

### LINKED ISSUE #478: Make reset button reset editor to initial state with a confirmation dialog

Tylopilus:
adds functionality to the reset button.

Closes #478


packages/monaco/src/EditorHooks.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react';
+
+export function useResetEditor(): { add: (cb: () => void) => void; dispatch: () => void } {
+  const ev = new CustomEvent('resetCode');
+
+  function add(cb: () => void) {
```
bautistaaa, line 6: i think a more appropriate name is subscribe.

out of scope but could probably make this more generic by being able to pass event name

packages/monaco/src/EditorHooks.ts
```diff
@@ -0,0 +1,20 @@
+import { useEffect } from 'react';
```
bautistaaa, line 1: we kebab case our file names but i'd argue you can just put this in the index file unless you see this growing

bautistaaa: oh i see its a separate entry point for importing

packages/monaco/src/EditorHooks.ts
Tylopilus: I agree with naming it subscribe but I question if we even should follow this or use zustand and create a store as we already have it as a dep. 

Depending on the other editor functionalities. But as these are just signals _I think_ a simple solution here is sufficient.

Tylopilus: I changed it to be more generic. I don't know if this is common to setup like this or if hooks being factories is more common.

PickleNik (approved):
lgtm

Issue #382, opened 2023-08-22
Need designs for profile page 


bautistaaa:
see #10

jhechtf:
@ernestoresende you think this is something you could do on the figma file or should I take a stab at it?
cc: @PickleNik

emalineg:
using max and nik's design to continue developing profile

emalineg:
passing this to max!

maxdemaio:
Desktop was done in #504! Now, we just need to work on mobile. Had some ideas I listed in the PR description of #504 and added some ideas to the Figma. If anyone wants to tackle that or pair lmk!

PickleNik:
> If anyone wants to tackle that or pair lmk!

![image](https://github.com/bautistaaa/typehero/assets/31113245/e42cf314-f2f4-44b1-a14b-0c481ceafad0)

maxdemaio:
Created a branch for this work here! https://github.com/bautistaaa/typehero/tree/profile-2

PR #504, opened 2023-08-29
Fix: Need designs for profile page 

maxdemaio:
This PR implements a vertical navigation for the profile page on desktop. On desktop, it will have two columns. See the changes in the attached video 👍🏻

Solves desktop for #382

https://github.com/bautistaaa/typehero/assets/44373521/51014450-1b43-4312-8e61-b71e1e2843a3

For now, on smaller screens it will wrap into one column. After discussion with @pizzaintheoven, we could match the media query on the 1/2 columns with `md:display hidden` so that those buttons are hidden and we could show a button to open a side navigation drawer (see https://ui.shadcn.com/docs/components/sheet). Also, in a future ticket we want to make that content span the whole height.


PickleNik (approved):
LLTM

PR #505, opened 2023-08-29
feat(web): highlight current path (explore, tracks)

QuiiBz:
Another very small PR to highlight the current URL in the nav:

Home:
Screenshot 2023-08-29 at 18 15 50

Explore:
Screenshot 2023-08-29 at 18 14 16

Tracks:
Screenshot 2023-08-29 at 18 14 13

### DIFF
--- a/apps/web/src/components/ui/navigation.tsx
+++ b/apps/web/src/components/ui/navigation.tsx
@@ -11,6 +11,7 @@ import {
   DropdownMenuTrigger,
 } from '@repo/ui';
 import { Loader2, LogIn, Moon, Plus, Settings, Settings2, Sun, User } from '@repo/ui/icons';
+import clsx from 'clsx';
 import { useTheme } from 'next-themes';
 import Link from 'next/link';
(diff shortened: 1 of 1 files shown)


apps/web/src/components/ui/navigation.tsx
```diff
@@ -80,14 +81,22 @@ export function Navigation() {
             </a>
             {featureFlags?.exploreButton ? (
               <Link href="/explore" className="ml-4">
-                <div className="hover:text-foreground text-foreground/80 transition-colors">
+                <div
+                  className={clsx('hover:text-foreground text-foreground/80 transition-colors', {
```
bautistaaa, line 85: isnt there an active prop we can use from `Link` or am i misremembering

bautistaaa (approved):

apps/web/src/components/ui/navigation.tsx
PickleNik: rewrite in Nuxt

PR #514, opened 2023-08-29
Bugfix: Hide Carousel arrows if the container cannot scroll anymore

andresmarpz:
This PR closes https://github.com/bautistaaa/typehero/issues/507


bautistaaa:
can you share a video of it working by chance

andresmarpz:
[video here](https://www.loom.com/share/bb6808e4e2b74f9c9cf613a9f8faa349). i didn't touch much about the code, mostly opted to show/hide with css since the event handlers are attached through the `useEffect` call, and that needs buttons to be mounted.

bautistaaa:
> [video here](https://www.loom.com/share/bb6808e4e2b74f9c9cf613a9f8faa349). i didn't touch much about the code, mostly opted to show/hide with css since the event handlers are attached through the `useEffect` call, and that needs buttons to be mounted.

what if there’s only one item. do both go away?

andresmarpz:
[improved here](https://www.loom.com/share/69851ea6f29f4e2cb491704ebef9fee8). That should do for different sizes of the container and/or the screen. The mobile version is untouched and basically hides the buttons since they can native scroll

Tylopilus (approved):
lgtm

bautistaaa:
> [improved here](https://www.loom.com/share/69851ea6f29f4e2cb491704ebef9fee8). That should do for different sizes of the container and/or the screen. The mobile version is untouched and basically hides the buttons since they can native scroll

@PickleNik 

its kind of weird to have the cards no align with the header. looks like a styling error when theres not enough items. what do you think?

PickleNik:
I agree but we need to somehow translate container width constraints into padding width for the carousel element

Rand0mLetters:
I'm not sure if this is related (or even worth working on), but I thought I'd drop this here

When there are no elements, a scrollbar appears anyway (idk where the code for this is, but I'd assume that it can be fixed with overflow: auto?)
image

PickleNik:
should we merge this now? @bautistaaa , I think adding left-padding to carousel container is a separate PR/issue

bautistaaa:
> should we merge this now? @bautistaaa , I think adding left-padding to carousel container is a separate PR/issue

sure. we can handle whatever issues are left after

apps/web/src/utils/useResizeObserver.ts
```diff
@@ -0,0 +1,43 @@
...
+      return () => {
+        resizeObserver.disconnect();
+      };
+    }
+  }, [ref, handleResize]);
+
+  return attached;
+};
+
+export default useResizeObserver;
```
bautistaaa, line 43: named export please

bautistaaa (approved):

apps/web/src/utils/useResizeObserver.ts
bautistaaa: ill just fix

PR #515, opened 2023-08-29
Fullscreen toggle

jproberson:
* On fullscreen toggle don't show left side of challenge
PR for https://github.com/bautistaaa/typehero/pull/515


apps/web/src/app/challenge/_components/challenge-layout.tsx
```diff
@@ -161,15 +165,19 @@ export function ChallengeLayout({ left, right }: ChallengeLayoutProps) {
...
-      <div
-        className="min-h-[318px] w-full overflow-hidden rounded-l-2xl rounded-r-xl border border-zinc-300 bg-white dark:border-zinc-700 dark:bg-zinc-800 lg:min-w-[500px]"
-        ref={leftSide}
-      >
-        {left}
-      </div>
-      <div className="resizer relative p-2" ref={resizer}>
-        <div className="absolute left-1/2 top-1/2 h-1 w-24 -translate-x-1/2 -translate-y-1/2 rounded-full bg-neutral-400 duration-300 group-hover:bg-neutral-600 group-active:bg-emerald-400 group-active:duration-75 dark:bg-neutral-700 group-hover:dark:bg-neutral-500 lg:h-24 lg:w-1" />
-      </div>
+      {!settings.isFullscreen && (
```
Tylopilus, line 168: Maybe we should reduce the width of leftpanel to 0 instead and keep the resize handle as this setting persists across page loads and not indicate visually that left panel is hidden

jproberson: I'm fine with it either way had a couple q's out in discord on the preferred method.

bautistaaa:
can you add a video of this if possible

jproberson:
Current Pr conditional render fullscreen:
fullscreen-conditional-render

Width to 0 fullscreen:
width-to-0

If we decide to go the width to 0 way, I'll probably want to include logic to reintroduce the min width style to the left side if the user attempts to resize it. Otherwise the top tabs are a little funky.

PickleNik (approved):
LGTM

PR #522, opened 2023-08-30
Kinda fixed the Spinner

ggrandi:
This fix is very much a patch (it just makes the button say open test cases)

It has unearthed that we need to change how we do the typechecks. Down the line, I think we should embed the typescript compiler and run it on submit.


bautistaaa:
does kinda mean its still breaking in some cases? 

if im understanding correctly i know we show a spinner on initialization so it can typecheck and show the appropriate squiggles. how would that work if we defer to clicking submit

ggrandi:
Kinda means that it is working now but the fix real fix would be that we fundamentally have to redo the logic for it.

My thoughts for doing this properly are to essentially run `tsc --noEmit` on the files in the browser so that we don't need to have the test cases panel open for them to work. I think this might also be necessary down the line if we want to add the the type benchmarking stuff

bautistaaa (approved):

packages/monaco/src/index.tsx
```diff
@@ -40,13 +40,14 @@ export function CodePanel(props: CodePanelProps) {
...
-  const [tsErrors, setTsErrors] = useState<TsErrors | undefined>(undefined);
+  const [tsErrors, setTsErrors] = useState<TsErrors>();
   const [localStorageCode, setLocalStorageCode] = useLocalStorage(
     `challenge-${props.challenge.id}`,
     '',
   );
 
-  const initialTypecheckDone = tsErrors === undefined;
+  const showSubmitSpinner = props.submissionDisabled || tsErrors === undefined;
+  console.log(props.submissionDisabled, tsErrors === undefined);
```
QuiiBz, line 50: To remove

QuiiBz (changes requested):

packages/monaco/src/index.tsx
bautistaaa: fixed

QuiiBz (approved):

PR #529, opened 2023-08-31
fix(admin): Re-adds rendering for comments 

jhechtf:
## Screen Shots

**BEFORE**

ss-before

**AFTER**
ss-after


apps/admin/package.json
```diff
@@ -19,7 +19,6 @@
     "@repo/auth": "workspace:*",
     "@repo/db": "workspace:*",
     "@repo/emails": "workspace:*",
-    "@repo/lint": "workspace:*",
```
bautistaaa, line 22: we dont need these?

jhechtf: Dog I don't know im just trying to get this to pass and I'd allow satan himself access to my code editor for it.

jhechtf: OK IDK why this is still showing up, it's still in there so i think that's a bug?

jhechtf:
@bautistaaa seems like everything short of the builds is passing now for reasons i don't know.

jhechtf:
screw it im gonna close this and re-do the work

PR #532, opened 2023-08-31
nix local development

ricardo-valero:
- flakes
- direnv
- devenv

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake . --impure
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,10 @@ yarn-error.log*
 
 #idea
 .idea
+
+# direnv
+.direnv
+
+# devenv
+.devenv
+result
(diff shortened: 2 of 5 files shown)


Tylopilus:
What is this for?

ricardo-valero:
This is tailored for Nix users and is limited to local development. 

Nix simplifies dependency management with reproducible builds, eliminating the need to install Node.js or MySQL directly on your local machine.

direnv facilitates the loading of the Nix environment, while devenv allows configuration of processes and services. Here, it’s used to add the MySQL service. 

Unlike Docker, which offers _repeatable_ builds, Nix not only ensures reproducibility but also provides isolation without the container overhead, resulting in a resource-friendly solution.

While I’m not advocating exclusive use, I wanted to share an alternative.

Recognizing that this might introduce additional elements to the repository, it offers an option for those interested.

This setup reflects how I personally configured my local development environment to begin exploring the repository.

bautistaaa:
im leaning towards closing this

thoughts @Hacksore @QuiiBz

Hacksore:
Yes nix is not worth supporting, pease close @bautistaaa

PR #534, opened 2023-08-31
newest new panel collapse

jproberson:
~* Go under 500? Go to 0~
~* Go over 10? Go to 500+~

Now it works by:
Goes down to 500, stops, then at 300 its collapses to 0
At 300 it'll go back to 500 and back up as you keep moving over.

Also double click will collapse / open as well.

### DIFF
--- a/apps/web/src/app/challenge/_components/challenge-layout.tsx
+++ b/apps/web/src/app/challenge/_components/challenge-layout.tsx
@@ -135,15 +135,13 @@ export function ChallengeLayout({ left, right }: ChallengeLayoutProps) {
     const mouseDownHandler = (e: MouseEvent | TouchEvent) => {
       if (e instanceof MouseEvent) {
         // Get the current mouse position
-        window.innerWidth > MOBILE_BREAKPOINT ? (x = e.clientX) : (y = e.clientY);
+        isDesktop ? (x = e.clientX) : (y = e.clientY);
       } else if (e instanceof TouchEvent) {
         // Get the current finger position
-        window.innerWidth > MOBILE_BREAKPOINT
-          ? (x = e.touches[0]?.clientX ?? 0)
-          : (y = e.touches[0]?.clientY ?? 0);
+        isDesktop ? (x = e.touches[0]?.clientX ?? 0) : (y = e.touches[0]?.clientY ?? 0);
       }
 
-      window.innerWidth > MOBILE_BREAKPOINT
+      isDesktop
         ? (leftWidth = leftSide.current?.getBoundingClientRect().width!)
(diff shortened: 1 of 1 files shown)


apps/web/src/app/challenge/_components/challenge-layout.tsx
```diff
@@ -80,9 +80,25 @@ export function ChallengeLayout({ left, right }: ChallengeLayoutProps) {
       const newLeftWidth = ((leftWidth + dx) * 100) / divideByW;
       const newTopHeight = ((topHeight + dy) * 100) / divideByH;
 
-      window.innerWidth > 1025
-        ? (leftRef.style.width = `${newLeftWidth}%`)
-        : (leftRef.style.height = `${newTopHeight}%`);
+      if (window.innerWidth > 1025) {
```
bautistaaa, line 83: lets avoid magic numbers and assign these to const vars

bautistaaa (changes requested):

apps/web/src/app/challenge/_components/challenge-layout.tsx
PickleNik: > lets avoid magic numbers and assign these to const vars

it's the `md:` screen size from https://github.com/bautistaaa/typehero/blob/9ec2e9c813c940df5f2387a1d355713ac16f8372/apps/web/tailwind.config.ts#L17C3-L17C22

bautistaaa: thats fine. just need to assign it to a meaningful name

jproberson:
@bautistaaa @PickleNik 
A bit of a rework tonight.

Now it works by:
Goes down to 500, stops, then at 300 its collapses to 0
At 300 it'll go back to 500 and back up as you keep moving over.

Also double click will collapse / open as well.

apps/web/src/app/challenge/_components/challenge-layout.tsx
Hacksore: @PickleNik does TW give you a way to import `lg` as a value from the theme in js land?

PickleNik:
> @bautistaaa @PickleNik A bit of a rework tonight.
> 
> Now it works by: Goes down to 500, stops, then at 300 its collapses to 0 At 300 it'll go back to 500 and back up as you keep moving over.
> 
> Also double click will collapse / open as well.

I don't like the behavior of having to drag back to 300 to un-collapse, I think un-collapsing should be instant as soon as u drag beyond 0, that's more intuitive I think

PickleNik (approved):
lgtm

Issue #525, opened 2023-08-30
toggle leftPanel width to 0


PickleNik:
toggle leftPanel width to 0 for maximum editing screen realestate

PickleNik:
being implemented in #523

PickleNik:
//todo: see [/challenge-layout.tsx] TODO for a failed attempt, replace it with working code from #523

Tylopilus:
thoughts on this?
https://jmp.sh/dLnD2c7E

PR #539, opened 2023-08-31
525 resize left panel

Tylopilus:
competing solution to #525

uses framer-motion instead of huge ass useEffect

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -58,6 +58,7 @@
     "react-markdown": "^8.0.7",
     "react-select": "^5.7.4",
     "react-syntax-highlighter": "^15.5.0",
+    "react-use-measure": "^2.1.1",
     "react-wrap-balancer": "^1.0.0",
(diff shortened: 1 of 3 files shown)


bautistaaa:
this is intriguing. can you post a video of it on both mobile and desktop. our feature flags have everything turned off on preview. need to get that sorted

Tylopilus:
@bautistaaa Here is the vid 
https://jmp.sh/rsr7ZbU4

pls help fixing lock conflict. how?

Hacksore:
so I was looking #534 by @jproberson which imo is a better UX and follows more what VSCode does.

I'm inclined to take that approach more but will let others comment what they think. cc @bautistaaa

https://github.com/typehero/typehero/assets/996134/82ac72fc-aaf8-4b0c-bc03-0c1599c135a4

PickleNik:
I do like the compactness and the smooth transition with framer, I hope there'll be a version of this after #534 is merged

PR #547, opened 2023-09-01
chore: adds sentry

bautistaaa:
### Note
it does not play nicely with vercel toolbar

### Errors
CleanShot 2023-09-01 at 00 04 31@2x


apps/web/sentry.client.config.ts
```diff
@@ -0,0 +1,30 @@
...
+// The config you add here will be used whenever a users loads a page in their browser.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from "@sentry/nextjs";
+
+Sentry.init({
+  dsn: "https://612a99a42def98a6f2951b66e0d9d776@o4505790239604736.ingest.sentry.io/4505790257627136",
+
+  // Adjust this value in production, or use tracesSampler for greater control
+  tracesSampleRate: 1,
```
lforst, line 11: I think this fails because the SDK attaches http headers to requests hitting the toolbar sever.

```suggestion
  tracesSampleRate: 1,
  
  // Request targets to which we want Sentry tracing headers attached
  tracePropagationTargets: ["localhost:3000", /^\//],
```

Hacksore: Well when I look at the toolbar network requests looks like they are sent to the following address.

`http://localhost:61987`, Can we just make sure sentry doesn't mess with that one? @bautistaaa

lforst: we can turn it into a regex

Hacksore: @lforst I tried what you suggested and it works, maybe the regex is not needed?

lforst: It is definitely not needed. It just seemed like you would have liked a "everything except" logic here, which tbh I don't fully recommend, as this option is opt-in by design.

You should put all targets inside of this option which you would like to trace. Anything that you don't own, or rather everything that you cannot control the CORS headers of should never be inside this array. Unfortunately the default value here is `["localhost", /^\//]` which matched `http://localhost:61987/`.

But I am glad this fixes it. We're gonna wait in the SDK to fix this until enough people run into this problem. Usage of `@vercel/toolbar` seems quite low tbg.

Hacksore: Makes sense. l I think we are the first people publicly using the toolbar on `localhost` 😬.

source: https://github.com/search?q=%40vercel%2Ftoolbar&type=code

bautistaaa: > Well when I look at the toolbar network requests looks like they are sent to the following address.
> 
> `http://localhost:61987`, Can we just make sure sentry doesn't mess with that one? @bautistaaa

i think its changes but unsure!

Hacksore (approved):

PR #549, opened 2023-09-01
chore: eslint format with prettier rules

QuiiBz:
Run Prettier rules when running ESLint, not sure if it was intended to not have it?

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -146,7 +146,7 @@ importers:
         version: 13.4.10(eslint@8.45.0)(typescript@5.1.6)
       eslint-plugin-import:
         specifier: ^2.27.5
-        version: 2.27.5(@typescript-eslint/parser@5.62.0)(eslint-import-resolver-typescript@3.5.5)(eslint@8.45.0)
+        version: 2.27.5(eslint@8.45.0)
       postcss:
         specifier: ^8.4.26
         version: 8.4.26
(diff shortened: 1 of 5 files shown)


Hacksore (approved):

QuiiBz:
Yeah let's wait for Anthony's answer

bautistaaa (approved):

Hacksore:
Don't merge until @anthonyshew weighs in.

jhechtf:
do we really have to use prettier though :c

Hacksore:
> do we really have to use prettier though :c

img

jhechtf:
> 

that does not answer any questions.

QuiiBz:
> that does not answer any questions.

Why wouldn't we want to use a formatter, without even having to think about it? It would avoid format errors in the CI and make sure we have good consistency in our codebase.

jhechtf:
Because I'm not a fan of most of the things that Prettier enforces and have found that consistent ESLint rules tend to remove the need for Prettier. 

And then people integrate Prettier into ESLint, which just seems like wearing socks on top of your socks to me.

Hacksore:
Let's close this for now and reopen if @anthonyshew can tell us if we should do this.

anthonyshew:
TL;DR: We don't need this PR. :)

> Pasting here from Discord!

eslint-plugin-prettier: Runs Prettier using ESLint which is slower than Prettier running Prettier. Additionally, we’d be running the Prettier formatting twice so doing unnecessary work.

eslint-config-prettier: Prevent running rules in ESLint that can overlap with your Prettier configuration. This is good, we hate when the formatter and the linter disagree and/or could potentially disagree.

Good news is: We don’t have to handle it ourselves because @vercel/style-guide already has eslint-config-prettier in it here: https://github.com/vercel/style-guide/blob/8ef6fc54e91092ab4288dee9588c2e70d8087322/package.json#L37

Issue #548, opened 2023-09-01
[Bug]: waitlist form throws error in dev


bautistaaa:
### What happened?

navigate to `/waitlist`

<img width="500" alt="CleanShot 2023-09-01 at 00 32 05@2x" src="https://github.com/bautistaaa/typehero/assets/3660667/95501a1f-eefa-4c85-af03-644e2407c5c0">

### What browsers are you seeing the problem on?

All

cosmiclasagnadev:
Think I have an idea about this. Would love to work on it @bautistaaa

Hacksore:
Fixed via #550

PR #550, opened 2023-09-01
548 - Fixed waitlist form error

cosmiclasagnadev:
Should remove the error from #548. 

Error might have been due to the fact that `{...field}` here passes down an extra ref that SelectPrimitive is not executing forwardRef on

image

Shadcn's documentation also doesn't pass down the rest of the `field` object when using Select component with react-hook-form inside a FormField

image

### DIFF
--- a/packages/email/yarn.lock
+++ b/packages/email/yarn.lock
@@ -170,12 +170,12 @@
   dependencies:
     eslint-visitor-keys "^3.3.0"
 
-"@eslint-community/regexpp@^4.4.0":
+"@eslint-community/regexpp@^4.4.0", "@eslint-community/regexpp@^4.6.1":
   version "4.8.0"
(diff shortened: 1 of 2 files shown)


packages/email/yarn.lock
Hacksore, line 1: Can you remove this pls 🙏

One day https://github.com/resendlabs/react-email/pull/892 be merged but today is not that day.

Hacksore (changes requested):

packages/email/yarn.lock
cosmiclasagnadev: Sure

cosmiclasagnadev: Should be good now @Hacksore

Sorry for the extra commit -- I cannot, for the life of me, remember how to remove the file without reverting back 🫠

Hacksore (approved):

PR #556, opened 2023-09-01
Fix: improve accessibility on landing (color/aria-labels)

maxdemaio:
This adds slightly higher contrast to the footer, landing description, and aria labels to our buttons in the navigation 👍🏻

(Note: performance is low only due to me being on dev server on local)
image


bautistaaa (approved):

maxdemaio:
:shipit:

PR #567, opened 2023-09-02
Fix comiter and authtor

Hacksore:


maxdemaio (approved):
Approving so we can test if it will show GH as the author.

PR #577, opened 2023-09-02
Chore/update ut

Hacksore:


apps/web/src/components/wizard/DescriptionEditor.tsx
```diff
@@ -24,6 +24,7 @@ export function DescriptionEditor({ form }: Props) {
                 <FormItem className="h-full">
                   <RichMarkdownEditor
                     dismissPreview
+                    allowImageUpload
```
Hacksore, line 27: @PickleNik this needs a new issue as it's kinda scuffed on this view:

image

PickleNik: @Hacksore true, pls make an issue, and for the stepper also

as for solutions: position: relative probably needs to be applied to the left side panel for the Uploadthing

and stepper titles need to be shortened on smaller screens

PR #586, opened 2023-09-02
challenge layout mobile panel

jproberson:
Some changes to make mobile left collapse work the same as desktop
Some refactoring to pull out the panel adjustment code to try and make it a little cleaner


is-it-ayush:
i'll start workin on #589 after this is merged cause i need the fancy changes 😃

bautistaaa:
this appear to fix all the previous issues I had. I think when its collapsed it should just be an expand button. Having to drag is odd

bautistaaa (approved):

PR #587, opened 2023-09-02
Added inline reply count

falcon-head:
Updated the comments.tsx file to include the toggle reply count button

https://github.com/typehero/typehero/assets/32938743/8ff99071-8939-4278-8d7f-8e3bffb5e692

@PickleNik


PickleNik (approved):
This is great! Thank you

PR #601, opened 2023-09-04
chore: remove default exports from non entry points

bautistaaa:
initial stab may have missed some but its better than none :)


jhechtf:
Not that I particularly mind, but why are we killing default imports?

PR #602, opened 2023-09-04
Ensure direct URL access redirects to the waitlist page

falcon-head:
image

image

Ensure that accessing the tracks page [https://typehero.dev/tracks] and challenges page [https://typehero.dev/challenge] redirects to the waitlist page when accessed directly.


Rand0mLetters:
Just /tracks and /challenges? what about /@username, /explore, and /wizard?

falcon-head:
> Just /tracks and /challenges? what about /@username, /explore, and /wizard?

Explore is already handled, and other pages display a 'you need to login' message. I can simply add the following code: if (pathname !== '/') { redirect(); } This will redirect all requests that are not from https://typehero.dev/ to the waitlist page. Just waiting for feedback.

Hacksore (approved):

PR #605, opened 2023-09-04
fix error dev not booting up

Tylopilus:
web:dev: [vercel-toolbar] No project info found. Make sure you run `vc link` in your project directory.
web:dev: `source` does not start with / for route {"source":"min/vs/(.*)","headers":[{"key":"Access-Control-Allow-Origin","value":"*"},{"key":"Access-Control-Allow-Methods","value":"GET"}]}
web:dev:
web:dev:
web:dev: Error: Invalid header found


jhechtf (approved):
🚢  it

PR #611, opened 2023-09-05
Lift comment state into url

Tylopilus:
This PR lifts the state of the show/hide comments window into the URL.

The advantage is that we can now directly link to the open comments window. This was presented by the view where the hover state indicates that comments / upvotes are clickable and provide their own action

Upon click on the comments button this will now leverage the opportunity and directly show the comments tab

https://jmp.sh/xWs4e6xD


bautistaaa:
this should be it’s own nested route like what we’re going to do for linkable comments.

challenge/[id]/comments

bautistaaa:
closing this in favor of deep linkable comments

Issue #506, opened 2023-08-29
[Bug]: The editor hover window doesn't show in the test cases


ggrandi:
<img width="922" alt="image" src="https://github.com/bautistaaa/typehero/assets/49359646/85df216c-332b-4a82-9bcb-4a19ec24b835">

I'm not sure what is causing it but it would be cool to see the hover menus, I know a similar thing happens in the typescript playground

oussamaBGZ:
i just created a PR for this.

bautistaaa:
moving panel to the bottom so non issue

PR #613, opened 2023-09-05
FIX: The editor hover window doesn't show in the test cases #506

oussamaBGZ:
i fixed the problem with the hover its just one line of code :) #506


bautistaaa:
we have a PR open to move it to the bottom current so this isnt needed!

bautistaaa:
actually this is probably helpful. will add in for now.

bautistaaa (approved):

bautistaaa:
can you run `pnpm checks:fix` youre failing prettier

oussamaBGZ:
@bautistaaa i had to fix some linting issues the ways "typeof import('monaco-editor')" was implemented caused a warning

packages/monaco/src/code-editor.tsx
```diff
@@ -5,6 +5,7 @@ import { useTheme } from 'next-themes';
 import { useMemo } from 'react';
 import { useEditorSettingsStore } from './settings-store';
 import { libSource } from './editor-types';
+import type * as MonacoEditor from 'monaco-editor';
```
bautistaaa, line 8: do you need the type keyword in addition to typeof of below?

oussamaBGZ: Technically we don't but Eslint will throw a warning:
"All imports in the declaration are only used as types. Use `import type`."

bautistaaa: ah ok it’s ok to eslint ignore when needed

bautistaaa (approved):

packages/monaco/src/code-editor.tsx
oussamaBGZ: ok done

PR #615, opened 2023-09-06
feat: linkable comments

bautistaaa:
this makes replies and comments deep linkable


bautistaaa:
> I think we can have the animations back and also the transparent background if we are using parallel routes / route interceptors like modals

animation was broken prior to this. not sure what happened but i’ll do the intercepting route here and see how that feels

apps/web/src/app/[locale]/challenge/_components/comments/comment.tsx
```diff
@@ -225,16 +255,39 @@ function SingleComment({
     }
   }
 
-  async function copyCommentUrlToClipboard() {
-    await navigator.clipboard.writeText(`${window.location.href}/${comment.id}`);
+  async function copyCommentUrlToClipboard(isReply: boolean) {
+    const commentId = isReply ? comment.parentId : comment.id;
+    await navigator.clipboard.writeText(
```
Comment (Rand0mLetters) [inline apps/web/src/app/[locale]/challenge/_components/comments/comment.tsx:266]: will this work if the user:
copies a comment url
goes to that url
copies another comment url?

see: https://github.com/Rand0mLetters/typehero/blob/8926155f945427da69aff0a5a58b5976e18cebea/apps/web/src/app/%5Blocale%5D/challenge/_components/comments/comment.tsx#L294-L311

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.tsx:266]: yes cause its using origin

PR #627, opened 2023-09-07
fix(admin): Add comment reports

jhechtf:
Re-adds comment report rendering, which was previously removed.
## Description
Makes some modifications to the admin package in order to re-add rendering for comment reports.

## Related Issue

N/A

## Motivation and Context

We need to be able to see the comments when viewing the comment reports
## How Has This Been Tested?

Locally, clicking through stuff to make sure.

image


apps/admin/src/app/challenge/[challengeId]/challenge-review.tsx
```diff
@@ -27,7 +27,7 @@ export function ChallengeReview({ challenge }: Props) {
         variant: 'success',
         title: 'Challenge Approved Successfully',
       });
-      router.push('/admin');
+      router.push('/');
```
Comment (bautistaaa) [inline apps/admin/src/app/challenge/[challengeId]/challenge-review.tsx:30]: do we need to append a querystring so it chooses the right tab?

apps/admin/src/app/report/[id]/actions.tsx
```diff
@@ -59,7 +59,7 @@ export function ReportActions({ report }: ReportActionsProps) {
         title: 'Success',
         description: <p>Comment deleted successfully</p>,
       });
-      router.push('/admin');
+      router.push('/');
```
Comment (bautistaaa) [inline apps/admin/src/app/report/[id]/actions.tsx:62]: do we need to append querystrings for correct tab?

packages/ui/src/index.ts
```diff
@@ -7,6 +7,8 @@ export * from './components/button';
 export * from './components/card';
 export * from './components/checkbox';
 export * from './components/dialog';
+export * from './components/data-table';
```
bautistaaa, line 10: we removed this on purpose. are you able to import directly?

jhechtf: It wasn't building when i tried to import it directly

apps/admin/src/app/report/[id]/actions.tsx
Comment (jhechtf) [reply apps/admin/src/app/report/[id]/actions.tsx:62]: mm probably.

bautistaaa:
rebase, barrel file is now deleted

apps/admin/src/app/report/[id]/actions.tsx
Comment (bautistaaa) [reply apps/admin/src/app/report/[id]/actions.tsx:62]: looks like so 

```ts
              <Link href="/?tab=tracks">
                <Button variant="ghost">Cancel</Button>
              </Link>
```

bautistaaa (approved):

PR #629, opened 2023-09-07
fix(TrackChallenge): Warning caused by conditional rendering of mock …

aakash19here:
## Description
There were some warnings related to the conditional rendering of mock challenges in the Track Challenge Card.
Therefore an alternate approach could be to use an immediately-invoked function expression. This allows you to avoid creating an entirely new external component for the mock condition. 

Here is the screenshot after the implementation was changed , resolving the warnings 
Before
Screenshot 2023-09-08 at 1 01 59 AM

After: 
Screenshot 2023-09-08 at 1 03 38 AM


apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
```diff
@@ -47,17 +47,22 @@ export function TrackChallenge({ challenge, className, mock }: TrackChallengePro
...
-              <input
-                className="peer hidden appearance-none"
-                type="checkbox"
-                id={challenge.id.toString()}
-              />
-              <div className="h-5 w-5 rounded-full border border-black/70 bg-black/10 duration-75 peer-checked:border-transparent peer-checked:bg-green-600/80 dark:border-white/50 dark:bg-white/10 peer-checked:dark:bg-green-300/80" />
-              <Check className="absolute left-1 my-auto h-3 w-3 scale-0 stroke-[4] text-white duration-300 peer-checked:scale-100 dark:text-black" />
-            </>
-          )}
+          {(() => {
```
Comment (QuiiBz) [inline apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:50]: That looks like a "hacky fix". We should probably extract everything to another component, so that Million can properly optimize it.

QuiiBz (dismissed):

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx]: Hey I have made the requested changes via making the return deterministic , according to https://million.dev/docs/rules

Screenshot 2023-09-08 at 6 48 06 PM

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
```diff
@@ -27,8 +31,17 @@ const BGS_BY_DIFFICULTY = {
   EXTREME: 'to-difficulty-extreme/20 dark:to-difficulty-extreme-dark/20',
 } as const;
 
-export function TrackChallenge({ challenge, className, mock }: TrackChallengeProps) {
+const MockComponent: FC<MockComponentProps> = ({ challenge }) => (
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:34]:
```suggestion
const MockComponent = ({ challenge }: MockComponentProps) => (
```

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
```diff
@@ -4,11 +4,15 @@ import { Check } from '@repo/ui/icons';
 import clsx from 'clsx';
 import { useIsMobile } from '~/utils/useIsMobile';
 import type { Tracks } from './track-grid';
+import type { FC } from 'react';
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:7]: remove

bautistaaa (approved):

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx]: Done , but I'm not able to pass the prettier check :(

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
```diff
@@ -47,18 +59,8 @@ export function TrackChallenge({ challenge, className, mock }: TrackChallengePro
...
-                className="peer hidden appearance-none"
-                type="checkbox"
-                id={challenge.id.toString()}
-              />
-              <div className="h-5 w-5 rounded-full border border-black/70 bg-black/10 duration-75 peer-checked:border-transparent peer-checked:bg-green-600/80 dark:border-white/50 dark:bg-white/10 peer-checked:dark:bg-green-300/80" />
-              <Check className="absolute left-1 my-auto h-3 w-3 scale-0 stroke-[4] text-white duration-300 peer-checked:scale-100 dark:text-black" />
-            </>
-          )}
-          {challenge.name}
+          <MockComponent challenge={challenge} />
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:62]: isnt this only supposed to show up when mock is true?

Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:62]: You are right Chris but conditioninally rendering the component makes it a un deterministic return 
https://million.dev/docs/rules#
Resulting warning and performance degradation

Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:62]: Can you please clarify why is the mock prop has to be passed in the components?

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
```diff
@@ -47,18 +59,8 @@ export function TrackChallenge({ challenge, className, mock }: TrackChallengePro
...
+          <div>{challenge.name}</div>
```
Comment (QuiiBz) [inline apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:63]: I don't think we need this extra div here

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
```diff
@@ -27,8 +30,17 @@ const BGS_BY_DIFFICULTY = {
   EXTREME: 'to-difficulty-extreme/20 dark:to-difficulty-extreme-dark/20',
 } as const;
 
-export function TrackChallenge({ challenge, className, mock }: TrackChallengeProps) {
+const MockComponent = ({ challenge }: MockComponentProps) => (
```
Comment (QuiiBz) [inline apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:33]: `MockComponent` shouldn't return this code unless `mock === true`, so we should also add a new `mock` prop to `MockComponentProps`

QuiiBz (changes requested):

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:33]: Sorry @QuiiBz I'm kind of a dumbass , can you tell me why is the mock prop required at the first place. Why can't it be always true ? Are we planning to replace the mock with something in future ?

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:63]: I'll remove it

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
Comment (QuiiBz) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:33]: In `feature-card.tsx` we pass a `mock` prop to be true, but in `track-card.tsx` this `mock` prop isn't used and so is false. We should keep this behavior to hide or show the checkbox

aakash19here:
Made the changes while keeping million JS happy and adding back the mock prop back, tell me what do you think @bautistaaa @QuiiBz :)

aakash19here:
removed div element  (cause I forgot it in previous commit ) :')

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx]: 👍🏻

apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx
```diff
@@ -27,8 +32,25 @@ const BGS_BY_DIFFICULTY = {
   EXTREME: 'to-difficulty-extreme/20 dark:to-difficulty-extreme-dark/20',
 } as const;
 
-export function TrackChallenge({ challenge, className, mock }: TrackChallengeProps) {
+const MockComponent = ({ challenge, mock }: MockComponentProps) => {
+  if (mock) {
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/tracks/_components/track-challenge-card.tsx:36]:
```suggestion
  if (!mock) {
    return null;
  }
  
  return ( ... )
```

bautistaaa (approved):
i dont really like what we have to do for this to make million happy but ill allow it for now

aakash19here:
Please someone review the latest commit , hopefully there isn't anything left to change @bautistaaa @QuiiBz

QuiiBz (approved):

PR #633, opened 2023-09-08
feat(web): static rendering on `/`

QuiiBz:
Enable static rendering only for `/`, so it doesn't break other pages that have server actions.


QuiiBz:
> `/waitlist` uses server actions. did you verify form submit still works?

Revert changes on the waitlist page, it only broke at runtime

PR #661, opened 2023-09-12
feat: allow solution comments to be shared - Closes #648

### LINKED ISSUE #648: [Feature]: Solution Comments should be linkable
### Description

Currently the challenge comments are linkable. The same implementation should be applied to solution comments

### Explanation

_No response_

zawarudo:
Allows sharing `Solution Comments` with the URL format
`http://localhost:3000/challenge/13/solutions/57/comments/175?replyId=176`

## Description

- Added `/challenge/[id]/solutions/[solutionId]/comments/[commentId]/page.tsx` to allow linking to `Solution Comments`

- Moved `<Comments />` outside of `<SolutionDetail />` to be able be used in 
`/challenge/[id]/solutions/[solutionId]/page.tsx` and
`/challenge/[id]/solutions/[solutionId]/comments/[commentId]/page.tsx`

- `<Comments />` will now be required to accept a `challengeId` , previously we were creating `Solution Comments` without having this saved in the DB under `Shared Solutions` data model.
This is because `rootId` can refer to a challenge Id when `<Comment type='CHALLENGE' />` but when [Comment type is 'SOLUTION' rootId refers to `rootSolutionId` and `rootChallengeId` is lost.](https://github.com/typehero/typehero/pull/661/files#diff-6ffcd17d428ccca4e0e17df4f046a75acc8e07b7448591d397466866fac97756L50)
But it is required to create a valid URL.
This could probably be cleaned up.

- created `getPreselectedSolutionCommentMetadata(solution, commentId)` server action in [getCommentRouteData.ts](https://github.com/typehero/typehero/compare/main...zawarudo:typehero:feature/linkable-solution-comments?expand=1#diff-695d0c4e63fb1bf65662842f1fd183848e316140649e9c64e6514b4dd8810108)

- Expanded type `PreselectedCommentMetadata` to allow the type of data from `getPreselectedSolutionCommentMetadata()`
Maybe this is not the best solution, I am still new to TS / Prisma

## Motivation and Context

https://github.com/typehero/typehero/issues/648

## How Has This Been Tested?

- Create a new comment on a submitted solution
- Create another comment replying to the comment above
- Hit share on both comments and visit the links,
- confirm we see the relevant comment with the highlight animation playing.
- Repeat steps under `Description` section of challenge to test comments there.

Visiting the top comment and its reply from the share function links.
`http://localhost:3000/challenge/13/solutions/57/comments/175`
`http://localhost:3000/challenge/13/solutions/57/comments/175?replyId=176`
image


apps/web/src/app/[locale]/challenge/[id]/comments/[commentId]/page.tsx
```diff
@@ -23,6 +23,7 @@ export default async function CommentPage({ params: { id, commentId } }: Props)
     <div className="relative h-full">
       <Description challenge={challenge} />
       <Comments
+        challengeId={Number(id)}
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/[id]/comments/[commentId]/page.tsx:26]: isnt this just the same value as `rootId` ?

Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/[id]/comments/[commentId]/page.tsx:26]: Correct, in this case,
but not for the `Solution Comment` path.

I felt the `rootId` introduced some confusion for me later in the code.
So making this data duplicate but explicit was a trade off I made.

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/[id]/comments/[commentId]/page.tsx:26]: AHA! good enough for now tbh

bautistaaa:
run `pnpm checks:fix`

bautistaaa:
did a test deploy of your branch and got this when i tried to share a solution comment

```
https://typehero-web-staging-git-fork-zawarudo-fea-82aa7e-trash-company.vercel.app/challenge/null/solutions/39/comments/152
```

zawarudo:
> did a test deploy of your branch and got this when i tried to share a solution comment
> 
> ```
> https://typehero-web-staging-git-fork-zawarudo-fea-82aa7e-trash-company.vercel.app/challenge/null/solutions/39/comments/152
> ```

Can you try creating a new comment and sharing it?
Sorry I may have forgotten to document clearly but previous comments weren't sending the challenge id to the server.

zawarudo:
Ran `pnpm checks:fix`

Please check the commit closely as I'm getting strange behavior with your project on my machine (windows non WSL)
- after running `pnpm i` it deletes some files which seems wonky

image

apps/web/src/app/[locale]/challenge/[id]/comments/[commentId]/page.tsx
Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/[id]/comments/[commentId]/page.tsx:26]: actually can’t you just grab challenge id from the url via useParams. that way you don’t need to save the challengeId or anything i think

Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/[id]/comments/[commentId]/page.tsx:26]: > actually can’t you just grab challenge id from the url via useParams. that way you don’t need to save the challengeId or anything i think
> 
> would also avoid having only new comments be linkable which is already sus imo

You're right, its cleaner to grab it from the params in `<Comments />`

I was hesitant to do this before, Because `<Comments />` lives in `_components/` and not `app/` typescript doesn't know what `params` will actually exist.

Will make the necessary changes and test, thanks!

bautistaaa:
> Ran `pnpm checks:fix`
> 
> Note: I'm getting strange behavior with your project on my machine (windows non WSL)
> 
> * after running `pnpm i` it deletes some files which seems wonky
> * If I include them back my whole project has errors as if there is no `node_modules`
> 
> I'd guess we should add these to `.gitignore`
> 
> image image

weird, i cant speak to windows issues personally since im on mac. im assuming youre running everything from root?

zawarudo:
> > Ran `pnpm checks:fix`
> > Note: I'm getting strange behavior with your project on my machine (windows non WSL)
> > 
> > * after running `pnpm i` it deletes some files which seems wonky
> > * If I include them back my whole project has errors as if there is no `node_modules`
> > 
> > I'd guess we should add these to `.gitignore`
> > image image
> 
> weird, i cant speak to windows issues personally since im on mac. im assuming youre running everything from root?

Yep running everything from root, no worries then.
If I find the root cause I'll PR it.

apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts
```diff
@@ -13,6 +13,7 @@ interface CommentToCreate {
   text: string;
   rootType: CommentRoot;
   rootId: number;
+  rootChallengeId: number;
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: do we need any changes to this file if we can just get the challenge id from the url?

Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: Whoops.

Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: We need this passed in when `rootId` refers to a `solution id`. `<Comment type="SOLUTION" />`

Are you saying we should grab the challenge id from the URL in this server file?
Because `useParams` is only available to client components.

We'd have to use fetch and grab it from the request instead of using a server function if we want to not pass this explicitly iirc.

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: challengeId is available in the server component via the prop (id)

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: oh wait… hmm okay

let me think about this for a moment

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: isn’t the url 

challenge/{id}/solution/{solutionId}

{id} should be reliable here?

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: i put the comment in the wrong line lol

apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts
```diff
@@ -47,9 +47,8 @@ export async function replyComment(comment: CommentToCreate, parentId: number) {
 
   const { rootId, ...commentToCreate } = {
     ...comment,
-    ...(comment.rootType === 'CHALLENGE'
-      ? { rootChallengeId: comment.rootId }
-      : { rootSolutionId: comment.rootId }),
+    rootChallengeId: comment.rootChallengeId,
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:50]: this part: why does this need to change?

apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts
Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: Ah got ya, I was following what you did before for normal comments (Where we save the challenge Id)

Note I did not change the schema at all, just following what I thought was your convention.

Since solutions are tied to challenges, it made sense for me to include the challenge Id as that's how the normal comments worked as well.

apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts
Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:50]: I thought this made things a bit more clear and explicit.

In the base case, normal comments work with `rootId = rootChallengeId`

Solution comments is kind of a special case
Where `rootId = rootSolutionId` and challengeId is lost in this situation.

I felt this change made that more explicit.

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:50]: the type is the discriminator here

apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts
Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: I guess we could avoid saving it if we read it from the component that handles generating the share URL.

But I felt since we're saving the challenge Id for normal comments, not saving it for other comments will be weird.

I guess you want to retroactively fix sharing for all the comments you already have without messing with the database?

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts:16]: solutions already have a relationship with a challenge which is why we don’t need that field to be populated. it’s redundant

Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts]: I think I understand, let me know if I'm on the right track with the latest changes.

https://github.com/typehero/typehero/pull/661/commits/9ff5ed18a0ea142e10348a2ef085d3a8109f241c

apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts
Comment (zawarudo) [reply apps/web/src/app/[locale]/challenge/_components/comments/comment.action.ts]: Reverted these changes and acting on your feedback.

zawarudo:
@bautistaaa 

When you follow a comment solution link that does not exist on the challenge, what would you like to happen?

[] Toast error
[] Redirect the user ie `/challenge/12/solutions/57/comments/180` => `/challenge/12/`

image

bautistaaa:
> @bautistaaa
> 
> When you follow a comment solution link that does not exist on the challenge, what would you like to happen?
> 
> [] Toast error [] Redirect the user ie `/challenge/12/solutions/57/comments/180` => `/challenge/12/`
> 
> image

hmmm this works

apps/web/src/app/[locale]/challenge/[id]/solutions/[solutionId]/comments/[commentId]/page.tsx
```diff
@@ -0,0 +1,37 @@
...
+    Number(solutionId),
+    Number(commentId),
+    Number(challengeId),
+  );
+
+  return (
+    <div className="relative h-full">
+      <SolutionDetails solution={solution} />
+      <Comments
+        rootId={solution.id!}
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/[id]/solutions/[solutionId]/comments/[commentId]/page.tsx:30]: use `solutionId` from the props

apps/web/src/app/[locale]/challenge/_components/comments/comment.tsx
```diff
@@ -261,15 +263,33 @@ function SingleComment({
...
-        isReply ? `?${searchParams.toString()}` : ''
-      }`,
-    );
+
+    const { rootType, rootSolutionId } = comment;
+    const baseURL = `${window.location.origin}/challenge/${challengeId}`;
+    const hasGetParams = isReply ? `?${searchParams.toString()}` : '';
+    let shareUrl;
+
+    switch (rootType) {
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/_components/comments/comment.tsx:276]: rootType can only be one or the other we can just make this more terse


```ts
const shareUrl = rootType ? ... : ...
```

bautistaaa (approved):

Issue #662, opened 2023-09-12
[Bug]: Resetting code does not reset the tests status


bautistaaa:
### What happened?

1. Solve a challenge.
2. Reset code.
3. Notice the tests are still passing

### What browsers are you seeing the problem on?

All

cosmiclasagnadev:
I can take a look at this @bautistaaa 🫡

bautistaaa:
issue still persists

bautistaaa:
fixed with: https://github.com/typehero/typehero/commit/71624b8405fedca2a7c03a8cb8254776d43b2b65

PR #666, opened 2023-09-13
fix: reset errors when resetting code as well

cosmiclasagnadev:
## Description
Added an effect for validating errors whenever user code is directly manipulated without going through editorState (e.g. reset code button)

## Related Issue
#662 

## Motivation and Context
To my knowledge, the validation of errors inside onMount and onChange don't do anything yet whenever `userCode` is directly manipulated (which is what `useResetEditor()` does).

https://github.com/typehero/typehero/assets/28593720/91fc7b0e-c129-40ec-a64f-dae36a7a6b6c


packages/monaco/src/split-editor.tsx
```diff
@@ -178,6 +178,50 @@ export default function SplitEditor({
...
+      }
+    }
+
+    return () => {
+      if (monacoAndEditorStateReady) {
+        typeCheck();
+        onMount?.tests?.(userEditorState, monaco);
+      }
+    };
+  }, [userCode, tests]);
```
bautistaaa, line 223: does monaco and userEditorState need to be added to this? not sure if anything else is missing

cosmiclasagnadev: re: monaco and userEditorState -- yeah onMount's tests are expecting 2 arguments 

Screenshot 2023-09-14 at 1 54 25 AM

re: triggering on other cases -- you mean if it validates the tests again on each input change? yah I think so


https://github.com/typehero/typehero/assets/28593720/e8748c50-d69a-439b-8294-bc223ea7ad26

bautistaaa: i just want to verify we're not running the typecheck code twice per keystroke or any other events and that the effect only triggers on the reset code case

bautistaaa: im asking if they needed to be added to the dep array because you are using them in your useEffect. im assuming you'd get a lint error so maybe its fine

cosmiclasagnadev: you're right, it is evaluated 2 times. let me work out some more fixes

bautistaaa: all good lmk when its ready!

cosmiclasagnadev: nevermind -- I was just dumb. this should be good now @bautistaaa

bautistaaa (approved):

packages/monaco/src/split-editor.tsx
```diff
@@ -177,6 +178,45 @@ export default function SplitEditor({
       }
     }
   }, [monaco]);
+  async function typeCheck() {
```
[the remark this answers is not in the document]
cosmiclasagnadev: should I put it in utils.ts in the same folder?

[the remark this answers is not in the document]
bautistaaa: sure! if youre able to do it easily!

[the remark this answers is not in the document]
cosmiclasagnadev: could I do it in a separate PR? not working properly yet when I move it to utils and might need to additional stuff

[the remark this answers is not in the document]
bautistaaa: yup! i said out of scope :)

Issue #669, opened 2023-09-14
[Bug]: Link input does not listen to all keystrokes, loses it's focus while typing after every character


aakash19here:
### What happened?

A bug happened!

https://github.com/typehero/typehero/assets/86520455/ce914c94-663b-4659-9e78-5422a6f7841d

### What browsers are you seeing the problem on?

Chrome

bautistaaa:
ahh another case of `million-ignore` search the codebase for it


you'll have to put it on the parent

aakash19here:
No , its not an issue with million ,its with the react-hook-form , its re-rendering the component after every keystroke and thus the focus is not persistent , I have solved it and will raise a PR shortly

aakash19here:
@bautistaaa review required

PR #670, opened 2023-09-14
fix(Settings): listening to all keystrokes while taking link input

aakash19here:
## Related Issue
https://github.com/typehero/typehero/issues/669

https://github.com/typehero/typehero/assets/86520455/b6c1bb12-2578-43b9-8d69-549a3438bef3

### DIFF
--- a/apps/web/src/components/settings/index.tsx
+++ b/apps/web/src/components/settings/index.tsx
@@ -44,6 +44,8 @@ export function Settings({ profileData, username }: { profileData: FormSchema; u
     register,
   } = form;
 
+  const { setFocus } = useForm();
+
   const userLinks = useWatch({ control: form.control, name: 'userLinks' });
 
   const { fields } = useFieldArray({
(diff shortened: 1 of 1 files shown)


apps/web/src/components/settings/index.tsx
```diff
@@ -105,6 +106,7 @@ export function Settings({ profileData, username }: { profileData: FormSchema; u
                       defaultValue={val.url}
                       placeholder="Link to social profile"
                       {...register(`userLinks.${i}.url`)}
+                      onChange={() => setFocus(`userLinks.${i}.url`, { shouldSelect: true })}
```
bautistaaa, line 109: you shouldnt have to setfocus on a keystroke

aakash19here: - Tried wrapping it with Controller component from react-hook-form , didn't worked their 
- Tried changing the key of FormItem component for not causing re-render , it didn't help either 
Can you please suggest something

bautistaaa:
fixed in main via million ignore

aakash19here:
Ohhh thanks trash <3

PR #674, opened 2023-09-15
chore: refactor type checking

amorriscode:
## Description

- [x] Moves `typeCheck` into `utils.ts` for shared functionality
- [x] Moves `subscribe` dependencies from root hook

## Motivation and Context

I moved the `deps` from `useResetEditor` to `subscribe` since they are only used to make sure the callback in that function isn't stale. 

## How Has This Been Tested?

Tested locally by clicking the reset button a bunch of times.

### DIFF
--- a/packages/monaco/src/editor-hooks.ts
+++ b/packages/monaco/src/editor-hooks.ts
@@ -8,11 +8,11 @@ class EditorEvents extends EventTarget {
 
 const event = new EditorEvents();
 type EventNames = 'resetCode';
-export function useResetEditor(deps: unknown[] = []): {
-  subscribe: (eventName: EventNames, cb: () => void) => void;
+export function useResetEditor(): {
+  subscribe: (eventName: EventNames, cb: () => void, deps: unknown[]) => void;
   dispatch: (eventName: EventNames) => void;
 } {
-  function subscribe(eventName: EventNames, cb: () => void) {
+  function subscribe(eventName: EventNames, cb: () => void, deps: unknown[] = []) {
     useEffect(() => {
       const fnc = () => {
         cb();
--- a/packages/monaco/src/editor-hooks.ts
+++ b/packages/monaco/src/editor-hooks.ts
@@ -9,7 +9,7 @@ class EditorEvents extends EventTarget {
 const event = new EditorEvents();
 type EventNames = 'resetCode';
 export function useResetEditor(): {
-  subscribe: (eventName: EventNames, cb: () => void, deps: unknown[]) => void;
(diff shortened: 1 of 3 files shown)


packages/monaco/src/editor-hooks.ts
```diff
@@ -8,11 +8,11 @@ class EditorEvents extends EventTarget {
...
 const event = new EditorEvents();
 type EventNames = 'resetCode';
-export function useResetEditor(deps: unknown[] = []): {
-  subscribe: (eventName: EventNames, cb: () => void) => void;
+export function useResetEditor(): {
+  subscribe: (eventName: EventNames, cb: () => void, deps: unknown[]) => void;
   dispatch: (eventName: EventNames) => void;
 } {
-  function subscribe(eventName: EventNames, cb: () => void) {
+  function subscribe(eventName: EventNames, cb: () => void, deps: unknown[] = []) {
```
amorriscode, line 15: Happy to undo this change but I think it makes more sense to have the dependencies scoped to `subscribe` since that's where they are used.

packages/monaco/src/split-editor.tsx
```diff
@@ -178,44 +179,17 @@ export default function SplitEditor({
...
-    }
-  }
-  useResetEditor([monaco, userEditorState]).subscribe('resetCode', () => {
-    if (monaco && userEditorState) {
-      typeCheck();
-      onMount?.tests?.(userEditorState, monaco);
-    }
-  });
+    },
+    [monaco, userEditorState],
```
bautistaaa, line 191: oh das better

packages/monaco/src/split-editor.tsx
```diff
@@ -62,6 +62,7 @@ export default function SplitEditor({
   monaco,
 }: SplitEditorProps) {
   const { settings, updateSettings } = useEditorSettingsStore();
+  const { subscribe } = useResetEditor();
```
amorriscode, line 65: Just a personal preference but this is a similar pattern used in other places where `useResetEditor` is used.

packages/monaco/src/editor-hooks.ts
bautistaaa: nope this is better. i def just rushed my solution in a couple mins lulz

bautistaaa (approved):

bautistaaa:
theres a typeerror

```
../../packages/monaco/src/index.tsx(61,20): error TS2554: Expected 3 arguments, but got 2.
```

probably cause you didnt make the deps an optional param

amorriscode:
@bautistaaa fixed. 😳

Issue #677, opened 2023-09-15
[Bug]: Waitlist Alert Icon Needs Padding


Jiansorge:
### What happened?

A bug happened! After joining the TypeHero waitlist, the "Thanks for signing up.." success alert does not correctly render the envelope svg icon on FireFox. It is overlapped with text and needs styling width.

-oldschoolrapper

![image](https://github.com/typehero/typehero/assets/38023093/56871aaf-1cbc-4b88-bc0e-06085cfcfc7a)

### What browsers are you seeing the problem on?

Firefox

### Relevant log output

```zsh
<div class="mt-3"><div class="relative w-full rounded-lg border p-4 [&amp;:has(svg)]:pl-11 [&amp;>svg+div]:translate-y-[-3px] [&amp;>svg]:absolute [&amp;>svg]:left-4 [&amp;>svg]:top-4 border-[#101b15]/50 dark:text-green-700 dark:border-[#101b15] [&amp;>svg]:text-green-300 bg-green-700 dark:bg-[#101b15]" role="alert"><svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round" class="h-4 w-4"><path d="M22 13V6a2 2 0 0 0-2-2H4a2 2 0 0 0-2 2v12c0 1.1.9 2 2 2h8"></path><path d="m22 7-8.97 5.7a1.94 1.94 0 0 1-2.06 0L2 7"></path><path d="m16 19 2 2 4-4"></path></svg><div class="text-sm [&amp;_p]:leading-relaxed text-left text-white">Thanks for signing up for the waitlist! Consider <a class="underline" href="https://twitter.com/typeheroapp" rel="noopener" target="_blank">following us on twitter</a> for general updates.</div></div></div>
```

heyanurag:
Hi folks! I'd like to work on this issue.

PR #684, opened 2023-09-16
fix: Alert icon overlapping with the text on Firefox

heyanurag:
## Description
:has selector is not supported in FF yet. - See https://github.com/tailwindlabs/tailwindcss/discussions/10206#discussioncomment-6604020

This removes the :has selector from the Alert component and adds padding-left to AlertDescription instead.

## Motivation and Context
#677 

## How Has This Been Tested?
Tested on Chrome, Brave and FF browsers.

Before:
image

After:
image

After on Chrome (No difference):
image


bautistaaa (commented):
there was an issue somewhere previously where someone added this but it broke other reolutions. can you make sure it works  as expected.

bautistaaa:
did you validate all resolutions here?

heyanurag:
> did you validate all resolutions here?

The Alert component was being used at one other place on the admin side.

Before:
image

After:
image


Here is a video for all resolutions. Looks good to me!

gif

bautistaaa:
awesome, looks like you have some eslint errors. then we can merge. its probably the ordering of your classes

`pnpm checks:fix` might fix it

bautistaaa (approved):

QuiiBz (approved):
Nice fix!

Issue #652, opened 2023-09-11
[Feature]: Add e2e tests


bautistaaa:
### Description

Not needed atm but soon.

Tired of things breaking

Options:
- cypress
- playwright
- other

billc-dev:
I would like to work on this. We started to port from Cypress to playwright and really liked it. It's a newer tool with less baggage compared to cypress. 

Parallel testing is built in and you don't need to pay for it unlike cypress. We didn't have to install any plugins either as everything just worked.

Features (all built-in):
- Parallelization
- Easy setup for browsers (chrome, firefox, webkit)
- HTML reporter with trace viewer
  - Trace viewer records console output and network requests (similar to chrome devtools)
- Code generation

oussamaBGZ:
> playwright

i think cypress is just fine for this project its very easy to setup with great doc and bigger community no need for a new tool i don't see the advantage, this is my opinion.

basokant:
+1 for Playwright, it's based

KaKashi1210:
echoing @oussamaBGZ, cypress setup is ez and one of the best docs out there

jhechtf:
It's another +1 from me for Playwright. Have used it in some other projects and liked the flow of it.

PR #685, opened 2023-09-17
feat: setup playwright and example tests

billc-dev:
## Description

- Playwright was setup on chrome with a few example tests for the home page, explore page, and challenge page
- A health check was added to make sure the database was ready after running `docker compose up -d --wait`
- The report is deployed to Cloudflare and the link will be on the Job summary page

The following Repository secrets are required:

`CLOUDFLARE_ACCOUNT_ID`
`CLOUDFLARE_API_TOKEN`
`EDGE_CONFIG`
`NEXTAUTH_SECRET`
`WEB_GITHUB_ID` (It is prefixed with `WEB_` because secrets starting with `GITHUB_` are reserved)
`WEB_GITHUB_SECRET`

## Related Issue

#652

## Motivation and Context

Demo report with traces: https://34221933.typehero.pages.dev

Click on a test and scroll down to traces and click on the **_image_** to open the trace viewer.
It records every step as a DOM snapshot with console logs and network requests.

Traces are set to record `on-first-retry`, so no traces are recorded when the test passes on the first try.
This would be the actual report when there are no errors https://73a09cf3.typehero.pages.dev

Deploying the report reduces the friction of viewing it.

I tried to deploy the report to Vercel, but I couldn't find a simple way to only upload static HTML files.
Cloudflare was used to deploy the report. If you don't have a Cloudflare account or if it's too much of a hassle, I can just DM you my `CLOUDFLARE_ACCOUNT_ID` and `CLOUDFLARE_API_TOKEN`. It's within the free tier.

## How Has This Been Tested?

The workflow has been tested on my fork
https://github.com/billc-dev/typehero/actions/runs/6210271610

https://github.com/typehero/typehero/assets/71155598/697a32f0-cf60-4bb0-be95-43190673679d


Hacksore:
This is awesome thanks getting this all setup!

We need to get the CF env vars and that's @bautistaaa next tasks so we can get this in. Let's setup a custom domain to like `e2e.typehero.dev` or something similar.

I've added the following secrets to the hub but still see the build failing for e2e 💀.
- [ ] CLOUDFLARE_ACCOUNT_ID
- [ ] CLOUDFLARE_API_TOKEN
- [x] EDGE_CONFIG
- [x] NEXTAUTH_SECRET
- [x] WEB_GITHUB_ID
- [x] WEB_GITHUB_SECRET

billc-dev:
You have to create a cloudflare pages project with a name of `typehero` and just `Create using direct upload` to initialize the project

Here is how to get `CLOUDFLARE_ACCOUNT_ID` and `CLOUDFLARE_API_TOKEN`
https://github.com/cloudflare/pages-action?tab=readme-ov-file#get-account-id

bautistaaa:
> This is awesome thanks getting this all setup!
> 
> We need to get the CF env vars and that's @bautistaaa next tasks so we can get this in. Let's setup a custom domain to like `e2e.typehero.dev` or something similar.
> 
> I've added the following secrets to the hub but still see the build failing for e2e 💀.
> 
> * [ ]  CLOUDFLARE_ACCOUNT_ID
> * [ ]  CLOUDFLARE_API_TOKEN
> * [x]  EDGE_CONFIG
> * [x]  NEXTAUTH_SECRET
> * [x]  WEB_GITHUB_ID
> * [x]  WEB_GITHUB_SECRET

can we just use staging?

bautistaaa:
> create a cloudflare pages

we have to create a cloudflare page? can we just hit our staging env?

bautistaaa:
added the account id and api token from the existing typehero.dev stuff that was there already. im not a CF user so lmk if i messed this up haha

billc-dev:
After some research, Github sets `Secret source: None` when the PR is opened from a fork for security measures. That's why the secrets are not present.
To get around this, we need to use `pull_request_target` but Github advises against this as it's possible to expose secrets:

> Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

billc-dev:
@bautistaaa I added a step to comment the report URL because the publishing workflow isn't shown in the checks and you'll need to dig through the actions tab to find the URL

bautistaaa:
> @bautistaaa I added a step to comment the report URL because the publishing workflow isn't shown in the checks and you'll need to dig through the actions tab to find the URL

im unable to locate the url

billc-dev:
@bautistaaa the `workflow_run` event will only be triggered if the workflow file is in the main branch
so the report isn't published for this PR but will in future ones 😀

bautistaaa (approved):

Issue #710, opened 2023-09-24
[Feature]: Users need the ability to unsubscribe from email list


bautistaaa:
### Description

add a link in email that will send them to an unsubcribe route

PR #715, opened 2023-09-24
feat: add unsubscribe for individuals on waitlist

bautistaaa:
## Description
We want to start sending updates to people on the waitlist as we near beta release. We expect some users to not like these emails so we want them to have a way to opt out.

- created a simple unsubscribe page
- added a link to the email that will send them to the unsubscribe page with prefilled input
- added header to email

## Related Issue
https://github.com/typehero/typehero/issues/710

## How Has This Been Tested?
tested locall

header added to email
CleanShot 2023-09-24 at 16 30 25@2x

CleanShot 2023-09-24 at 16 31 17@2x


Hacksore (approved):

apps/web/src/app/[locale]/(email)/unsubscribe/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+export type FormSchema = z.infer<typeof formSchema>;
+export default function UnsubscribePage(props: { searchParams: { email?: string } }) {
+  const { toast } = useToast();
+
+  const form = useForm<FormSchema>({
+    resolver: zodResolver(formSchema),
+    defaultValues: {
+      email: props.searchParams.email ?? '',
+    },
+  });
```
Comment (ayrock-dev) [inline apps/web/src/app/[locale]/(email)/unsubscribe/page.tsx:33]: FFT re: privacy policy, `?email=` as a query param will show up in any logs for page hits. this means sending the user's email to a service like Sentry, for example, anywhere we log the full request url

Comment (bautistaaa) [reply apps/web/src/app/[locale]/(email)/unsubscribe/page.tsx:33]: hmm is that an actual concern for sentry? whats the usual workaround, encryption?

Comment (ayrock-dev) [reply apps/web/src/app/[locale]/(email)/unsubscribe/page.tsx:33]: :shrug: maybe not a concern, just a note

Comment (bautistaaa) [reply apps/web/src/app/[locale]/(email)/unsubscribe/page.tsx:33]: nah this is valid and now im scared lollll

PR #717, opened 2023-09-25
feat: add in-progress challenges to profiles

### LINKED ISSUE #714: [Feature]: Add link to in-progress problems
### Description
https://github.com/typehero/typehero/assets/53154523/4e4b9b0b-7199-4cd2-9bb8-ce69f339658b

It would be nice to have something similar to the "Solutions" panel, where you can click on the problems you have solved. Right now, "In-Progress" panel names the challenge, but doesn't let you click on it to go to the problem!

### Explanation

This would be useful because I want to easily be able to go back to problems I have started, but not finished yet.

amorriscode:
## Description

- [x] Resolves #714
- [x] Actually fetches in-progress challenges and displays them
- [x] Changes "Solutions" to "Submissions" on the progress
- [x] Changes the way data is fetched on the profile (fetch less data)
- [x] Adds descriptions to the "In-Progress"/"Submissions" pages to match other settings pages
- [x] Adds coloured badge to the "Submissions" tab

## Related Issue

#714 

## Motivation and Context

I went in to resolve #714 but then realized that the "In-Progress" page was actually hard-coded. I moved some of the data fetching logic around so the "In-Progress" and "Submissions" would use similar patterns.

I made an executive decision and renamed "Solutions" to "Submissions" because a solution on a challenge is a submission that was successful whereas this page shows all (25 for now) of a user's submissions.

## How Has This Been Tested?

With 1337 skillz (manual clicking around, boatloads of `console.log`, etc.)

Screenshot 2023-09-24 at 7 49 53 PM

Screenshot 2023-09-24 at 7 49 59 PM


apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx
```diff
@@ -1,39 +1,55 @@
...
+      {
+        createdAt: 'desc',
+      },
+    ],
+    take: 25,
+    include: {
+      challenge: true,
+    },
+    distinct: 'challengeId',
+  });
```
Comment (odicho) [inline apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx]: I don't think this logic is correct. By definition, "in-progress" means something that has been started but not completed yet. Your query is just returning a list of submissions where the user has has made an attempt and failed previously, but doesn't take into account the fact whether the user might have actually also solved it in a different attempt.

For example if the submissions table looked like this (simplified):
```typescript
[
  { id: 1, challengeId: 1, userId: 1, isSuccesful: false}, 
  { id: 2, challengeId: 1, userId: 1, isSuccesful: true}
]
```

As you can see, user 1 attempted challenge 1 and failed. Challenge 1 is now in-progress. User 1 made another attempt and submitted a correct solution and the submission is successful. Challenge 1 is now completed.

For this example, your query will return challenge 1 still because it has `isSuccessful: false`.

There are multiple ways to solve this efficiently with Prisma, but the general question is: Find all challenges that have at least one or more submissions AND none of them are successful.

I might be wrong with the syntax, but I think this query (or something similar) should work:
```typescript
return await prisma.challenge.findMany({
  where: {
    submissions: {
      none: {
        where: {
          userId: id,
          isSuccessful: true,
        },
      },
    },
  },
});
```

Read more about Prisma relation filters [here](https://www.prisma.io/docs/concepts/components/prisma-client/relation-queries#relation-filters)

bautistaaa:
sorry i just guinea pigged your PR with unstable cache

apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx
```diff
@@ -1,39 +1,55 @@
...
+});
+
+export async function InProgressTab({ userId }: { userId: string }) {
+  const submissions = await getInProgressSubmissions(userId);
```
Comment (odicho) [inline apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx]: I think semantically, it is better to name this variable `challenges` because we want to display a list of challenges that are **in-progress**!

apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx
Comment (odicho) [reply apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx]: Again, I might be wrong with Prisma syntax (a bit rusty), but I think a nested relationship query can be something like this?

```typescript
return await prisma.challenge.findMany({
  where: {
    submissions: {
      none: {
        where: {
          userId: id,
          isSuccessful: true,
        },
      },
    },
  },
  select: {
    id: true,
    submissions: {
      orderBy: {
        createdAt: 'desc'
      },
      take: 1,
      select: {
        createdAt: true
      }
    }
  }
});
```

This query will return a list of challenges that are in-progress for the current user, and also return the createdAt field of the latest failed submission.

apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx
```diff
@@ -1,39 +1,55 @@
...
-            <TableCell className="font-medium">{challengeTitle.challengeTitle}</TableCell>
-            <TableCell>{challengeTitle.challengeStatus}</TableCell>
+        {submissions.map((submission) => (
+          <TableRow key={submission.id}>
+            <TableCell className="font-medium underline">
+              <Link href={`/challenge/${submission.challenge.id}`}>
+                {submission.challenge.name}
+              </Link>
+            </TableCell>
+            <TableCell>{getRelativeTime(submission.createdAt)}</TableCell>
```
Comment (odicho) [inline apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx:37]: As you can see, you're iterating through `submissions` but you're technically listing out `challenges` instead, so I definitely think it is a good idea to get a list of challenges instead as suggested above!

odicho:
Great work @amorriscode! 

I added a couple of comments for you, let me know what you think.

In addition to that, I would definitely run the "Submissions" vs "Solutions" with Chris to double check the actual intention. If you change it to submissions, it should list out all submissions like you said (which is what the code is doing at the moment). I am thinking that the intention was for it to show a list of solved challenges, but the code just implemented listing out all submissions...? Not sure.

Solutions, on the other hand, indicates either correct submissions OR if there is ever a tab like Leetcode, where you can post solutions in the discussions or whatever, then it can mean that.

amorriscode:
@odicho I adjusted the challenge query. I think it's kind of gross querying this way but Prisma doesn't have proper nested queries without using `rawQuery` so I went with it.

@bautistaaa two things for you:
1. Is it okay to change "Solutions" to "Submissions" as I have here? You can see @odicho's comment and my explanation as to why I did it above.
2. With `withUnstableCache` I noticed it was actually messing up the queries. You had the cache key as `undefined` but I think we'd want to set that? I made that change but I'm not sure if it's actually correct.

apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx
```diff
@@ -1,42 +1,80 @@
...
+        {
+          submission: {
+            none: {
+              userId: id,
+              isSuccessful: true,
+            },
+          },
+        },
+        // Make sure there is at least one submission
+        { submission: { some: { userId: id, isSuccessful: false } } },
```
Comment (odicho) [inline apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx:57]: Is this needed?

The none part makes sure there is either no successful submissions, or no submissions at all. I'm not sure what value the AND adds here.

Comment (amorriscode) [reply apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx:57]: Without the `AND` it returned all challenges that had no submissions by my user.

Comment (odicho) [reply apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx:57]: Oh, you're totally right! Ugh, the Prisma syntax is such a pain to read 😂

Comment (amorriscode) [reply apps/web/src/app/[locale]/[username]/_components/dashboard/in-progress-tab.tsx:57]: Yeah I was disappointed we couldn't just use a subquery for this. I guess we could just do `rawQuery` but this works too. 😛

bautistaaa:
> @odicho I adjusted the challenge query. I think it's kind of gross querying this way but Prisma doesn't have proper nested queries without using `rawQuery` so I went with it.
> 
> @bautistaaa two things for you:
> 
> 1. Is it okay to change "Solutions" to "Submissions" as I have here? You can see @odicho's comment and my explanation as to why I did it above.
> 2. With `withUnstableCache` I noticed it was actually messing up the queries. You had the cache key as `undefined` but I think we'd want to set that? I made that change but I'm not sure if it's actually correct.
> 
> Also I thought the screenshots would be okay because it's pretty static but here is a video:
> 
>  Screen.Recording.2023-09-25.at.8.56.27.AM.mov 
> We'll want to invalidate the cache when folks submit attempts to challenges but I figure that can be done in a different PR.

in unstable_cache the `queryKey` is the `tags`. can you elaborate how it was broken?

amorriscode:
> can you elaborate how it was broken?
> 
> in unstable_cache the `tags` are how we invalidate via `revalidateTags`

When I was switching tabs on the profile the query had the same results. This is the code I saw on the [Next.js docs](https://nextjs.org/docs/app/building-your-application/caching#unstable_cache):

```react
import { unstable_cache } from 'next/cache'
 
export default async function Page() {
  const cachedData = await unstable_cache(
    async () => {
      const data = await db.query('...')
      return data
    },
    ['cache-key'],
    {
      tags: ['a', 'b', 'c'],
      revalidate: 10,
    }
  )()
}
```

I've never used this API but wouldn't we want the `cache-key` to be unique per query?

bautistaaa:
> > can you elaborate how it was broken?
> > in unstable_cache the `tags` are how we invalidate via `revalidateTags`
> 
> When I was switching tabs on the profile the two queries had the same results. This is the code I saw on the [Next.js docs](https://nextjs.org/docs/app/building-your-application/caching#unstable_cache):
> 
> ```
> import { unstable_cache } from 'next/cache'
>  
> export default async function Page() {
>   const cachedData = await unstable_cache(
>     async () => {
>       const data = await db.query('...')
>       return data
>     },
>     ['cache-key'],
>     {
>       tags: ['a', 'b', 'c'],
>       revalidate: 10,
>     }
>   )()
> }
> ```
> 
> I've never used this API and the documentation is sparse. Wouldn't we want the `cache-key` to be unique per query? Happy to undo this particular bit if it isn't correct!

the interesting part here is that you can only currently remove the cache by `tag` and not `key` which basically makes that field useless atm until they update their docs

in the meantime its probably best to make the tag `${userId}-submissions` or something specific

bautistaaa:
we can merge this for now and ill play around with unstable cache later

bautistaaa (approved):

Issue #711, opened 2023-09-24
[Feature]: load all replies when panel is expanded with load more behavior


bautistaaa:
### Description

related: https://github.com/typehero/typehero/issues/708

when replies are expanded load all the replies in memory but paginate on the client.

see this for example: https://leetcode.com/problems/two-sum/description/comments/1565865

oreoluwa-bs:
can I work on this issue?

bautistaaa:
> can I work on this issue?

sure

bautistaaa:
be mindful of the individual working on the other related task

oreoluwa-bs:
Alright thank you

PR #718, opened 2023-09-25
feat: load all replies when panel is expanded with load more behavior

oreoluwa-bs:
## Description

- Fetches all comments(replies) stores that in react query cache
- Added client side pagination for replies which is managed by `useInfiniteQuery`

## Related Issue

#711 

## Motivation and Context

#711

## How Has This Been Tested?

Manual clicking through the comments panel

https://github.com/typehero/typehero/assets/46888096/f93dbfa7-7627-40a7-9dfc-f53886d15e11


bautistaaa:
looks great so far!

found a small bug

- if you create a new comment in the replies it does not appear(however the count updates). the new reply should append to the bottom

oreoluwa-bs:
Thank you.
Had issues with invalidating the query. It took too long for the state to update so I opted for refetch interval instead.

bautistaaa:
> Thank you. Had issues with invalidating the query. It took too long for the state to update so I opted for refetch interval instead.

this is broken for me

CleanShot 2023-09-26 at 21 51 36

bautistaaa:
i was thinking you could probably just do optimistic updates or something

https://tanstack.com/query/v4/docs/react/guides/optimistic-updates

oreoluwa-bs:
I thought about that also but it was also broken when you delete a comment

bautistaaa:
the useeffect just spams the network tab now

ill take a stab at this as well

oreoluwa-bs:
I realised it was because data was part of the dependency array

oreoluwa-bs:
It's no longer spamming the network tab and changes are pagination is fully on the client

bautistaaa (approved):

PR #745, opened 2023-10-01
migrate emails to mailchimp

### LINKED ISSUE #741: [Feature]: Move to mailchimp for marketing / waitlist emails
### Description

- export current waitlist to mailchimp
- redo waitlist form to use mailchimp
- remove waitlist table

note: we can stay on sendgrid for transactional emails still that can be managed in the new user settings

### Explanation

_No response_

bautistaaa:
## Description
Resend is not a great candidate for bulk email so exporting our current waitlist to mailchimp.
- Keeping the old server action so we can repurpose it for transactional things later

## Related Issue

Closes #741

## How Has This Been Tested?
locally


bautistaaa:
pushing so i can test it in prod

PR #769, opened 2023-10-04
chore: remove waitlist table

bautistaaa:
## Description
We migrated to mailchimp so we no longer need the waitlist table or the unsubscribe form as this will be managed in settings later


ayrock-dev:
to doc ur process here, do u run `prisma db push` on prod at some point after merging this with ur prod planetscale creds?

bautistaaa:
> to doc ur process here, do u run `prisma db push` on prod at some point after merging this with ur prod planetscale creds?

ideally yes but since we have no data i just run pnpm refresh

jhechtf (approved):
LGTM

Hacksore (changes requested):
wait but like, we would still have the waitlist button on the home page.

Remove it?

bautistaaa:
> wait but like, we would still have the waitlist button on the home page.
> 
> Remove it?

everything looks the same we just use mail chimp. they have their own user mgmt so we don’t need unsubscribe

Hacksore (approved):

Issue #752, opened 2023-10-02
[Feature]: Ability to pin solutions


bautistaaa:
### Description

like leetcode want to be able to pin solutions for a challenge. see

shohan2001:
Can you please assign me this issue?

bautistaaa:
> Can you please assign me this issue?

sorry ill be taking this one

PR #774, opened 2023-10-05
feat: ability to pin solutions

bautistaaa:
## Description
Adds the ability to pin solutions.

## Related Issue

Closes https://github.com/typehero/typehero/issues/752

## Motivation and Context
We want the ability to pin solutions or anything for that matter. For the time being it will just be solutions and ideally it would be solutions by the official typehero account. The use case may never come up but its a nice feature to have in the case we want to highlight great solutions.

## How Has This Been Tested?
Locally, trust me bro

Badge that indicates a solution is pinned
CleanShot 2023-10-04 at 22 48 46@2x

This pin is only available for mods/admins
CleanShot 2023-10-04 at 22 49 11@2x

### DIFF
--- a/apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts
+++ b/apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts
@@ -1,5 +1,6 @@
 'use server';
 import { prisma } from '@repo/db';
+import { revalidateTag } from 'next/cache';
 
 interface Args {
   challengeId: number;
@@ -17,4 +18,12 @@ export async function postSolution({ challengeId, description, title, userId }: 
       userId,
     },
   });
+}
+
+export async function pinOrUnpinSolution(id: number, isPinned: boolean) {
+  await prisma.sharedSolution.update({
+    where: { id },
+    data: { isPinned },
+  });
+  revalidateTag(`challenge-${id}-submissions`);
 }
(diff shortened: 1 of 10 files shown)


apps/web/src/app/[locale]/challenge/[id]/solutions/_components/solution-detail.tsx
```diff
@@ -1,26 +1,36 @@
...
   solution: ChallengeSolution;
 }
 export function SolutionDetails({ solution }: Props) {
   const { data: session } = useSession();
+
+  const showPin = [RoleTypes.ADMIN, RoleTypes.MODERATOR].some(
+    (i) => session?.user.role.includes(i),
+  );
+  const handlePinClick = async () => {
+    await pinOrUnpinSolution(solution.id, !solution.isPinned);
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/[id]/solutions/_components/solution-detail.tsx:32]: didnt debounce or optimistic update since this is an admin feature

apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts
```diff
@@ -18,3 +19,11 @@ export async function postSolution({ challengeId, description, title, userId }:
     },
   });
 }
+
+export async function pinOrUnpinSolution(id: number, isPinned: boolean) {
+  await prisma.sharedSolution.update({
```
Comment (Hacksore) [inline apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts:24]: add server side validation that the homie is admin plox

Hacksore (changes requested):

apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts
Comment (Hacksore) [reply apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts:24]: @bautistaaa we can let this go as long as you promise to make a ticket for handling this repo wide?

Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts:24]: nah ill fix tonight!

Hacksore (approved):

PR #783, opened 2023-10-05
fix: Dark mode profile bio wrong theme colors

### LINKED ISSUE #772: [Bug]: Dark mode profile bio wrong theme colors
### What happened?

When you first pull up to the profile in dark mode you see this but if you toggle to light mode then back to dark it goes away.

Notice the bio field has:
- wrong box shadow color
- text color is not visible

image

_No response_

### Relevant log output

_No response_

MrVSiK:
`system` or `undefined` values of theme were not being handled appropriately. Added a switch case to map possible theme values to `data-color-mode` values.

## Description
As per [next-theme docs](https://www.npmjs.com/package/next-themes#API) the `defaultTheme` attribute is set to the value `system`. While setting `data-color-mode` in RichMarkdownEditor, this causes it to default to `light` theme for the very first time (when `theme` in localStorage is set to `system`). After switching themes manually, the `theme` value in localStorage resets to either `dark` or `light` causing no further problems.

I resolved it by creating a switch block for setting `data-color-mode` based on `theme` and handling the default case. 

Alternate method: Enforcing a application wide default theme by setting `defaultTheme` attribute in `ThemeProvider`

## Related Issue

Closes #772

## Motivation and Context

#772 

## How Has This Been Tested?

Manually tested the following conditions:
1. `theme` in localStorage is `system`
2. `theme` in localStorage is `light`
3. `theme` in localStorage is `dark`

## Screenshots/Video (if applicable):'
System
Dark
Light

### DIFF
--- a/apps/web/src/components/rich-markdown-editor.tsx
+++ b/apps/web/src/components/rich-markdown-editor.tsx
@@ -38,9 +38,22 @@ export function RichMarkdownEditor({
 
   const { theme } = useTheme();
   useEffect(() => {
-    theme === 'dark'
-      ? document.documentElement.setAttribute('data-color-mode', 'dark')
-      : document.documentElement.setAttribute('data-color-mode', 'light');
+    switch(theme){
+      case 'dark': {
+        document.documentElement.setAttribute('data-color-mode', 'dark');
+        break;
+      }
+
+      case 'light': {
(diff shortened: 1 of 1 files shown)


Hacksore (changes requested):
@MrVSiK can you run `pnpm checks:fix` to clean up the formatting?

Hacksore (approved):

PR #799, opened 2023-10-07
Bump prisma

Hacksore:
attempt to fix the errors we see in staging below by bumping prisma version.

```
web:build: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason:
20:03:33.636 | web:build: PrismaClientInitializationError: Unable to require(`/vercel/path0/node_modules/.prisma/client/libquery_engine-rhel-openssl-1.0.x.so.node`).
20:03:33.636
```


Hacksore:
we will do this later it's hard rn

Issue #777, opened 2023-10-05
[Feature]: cookie consent banner


Hacksore:
### Description

make goodlookin™ based on design from @PickleNik if he can otherwise use mine.

but for just just a banner and use `localStorge` to set if they've seen it or not.

![image](https://github.com/typehero/typehero/assets/996134/c673cd97-d8f1-49be-ba8e-6a07b6a63563)

G-R3:
can I hack away at this?

Hacksore:
@G-R3 :shipit:

G-R3:
hey @Hacksore what would you like the text content to say? Also does it make sense to just have one button instead of the two? maybe just have it say `Understand` or something 😅. I did some work on this and this what I have so far.

https://github.com/typehero/typehero/assets/48399420/af9737d8-4a71-4213-80d3-d13dfadcac0e

<img width="1511" alt="Screenshot 2023-10-06 at 12 15 14 AM" src="https://github.com/typehero/typehero/assets/48399420/0dc4fc47-d148-4277-b891-c894d303ec25">
<img width="1509" alt="Screenshot 2023-10-06 at 12 15 02 AM" src="https://github.com/typehero/typehero/assets/48399420/57d64705-ca1f-4623-89d2-0ba826fed85d">
<img width="354" alt="Screenshot 2023-10-06 at 12 14 47 AM" src="https://github.com/typehero/typehero/assets/48399420/a755100d-6dac-4baf-a4de-23736c67ad0b">
<img width="355" alt="Screenshot 2023-10-06 at 12 14 06 AM" src="https://github.com/typehero/typehero/assets/48399420/bae35682-53bb-4948-a98b-419780d27bb6">

PR #800, opened 2023-10-07
feat: cookie consent banner

G-R3:
## Description

Introduces a cookie consent banner 

## Related Issue

Closes https://github.com/typehero/typehero/issues/777

https://github.com/typehero/typehero/assets/48399420/1f89ab3f-ff33-40b2-b513-4fa867a122cd

Screenshot 2023-10-06 at 12 15 14 AM
Screenshot 2023-10-06 at 12 15 02 AM
Screenshot 2023-10-06 at 12 14 47 AM
Screenshot 2023-10-06 at 12 14 06 AM

### DIFF
--- a/apps/web/src/app/layout.tsx
+++ b/apps/web/src/app/layout.tsx
@@ -8,6 +8,7 @@ import { Toaster } from '@repo/ui/components/toaster';
 import { Providers } from './providers';
 import { buildMetaForDefault } from './metadata';
 import { Navigation } from '~/components/navigation';
+import { CookieBanner } from '~/components/cookie-banner';
 
 const inter = Inter({ subsets: ['latin'] });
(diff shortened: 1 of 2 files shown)


G-R3:
copying a comment i left on the issue here
> hey @bautistaaa @Hacksore what would you like the text content to say? Also does it make sense to just have one button instead of the two? maybe just have it say I understand or something 😅. I did some work on this and this what I have so far.

Hacksore:
Perhaps it's a good idea to make a new page explaining all the cookies we use but that can be done later and not a blocker here.

Right now I think it's these:
- session
- locale (not really used yet)
- 
image

I think what you have now is fine.

Hacksore (approved):

PR #818, opened 2023-10-09
feat: authors can delete their solution

### LINKED ISSUE #804: [Feature]: Authors of a solution should be able to delete them
### Description

Just like an author can delete their own comment, they should be able to delete their solution.

Add a delete button next to the others
CleanShot 2023-10-08 at 01 00 01@2x

### Explanation

_No response_

zaCKoZAck0:
## Description

1. Added a delete button to let authors delete their solutions
image
2. Delete button is not visible if,
2.1. User is not signed in
2.2. User is not author of the solution
image
3. Added auth guard that only author can delete solutions
4. Added modal inspired from delete comment modal, with adjustments according to solutions use case. 
image
4.1. Added scrolling to markdown preview for solution, so long solutions don't break modal.
## Related Issue

Closes #804 


apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts
```diff
@@ -26,6 +26,21 @@ export async function postSolution({ challengeId, description, title, userId }:
...
+    where: { id }
+  })
+  if (!isAuthor(session, solution?.userId)) {
+    throw new Error('Only author can delete their solution.')
+  }
+
+  await prisma.sharedSolution.delete({
+    where: { id }
+  })
+  revalidateTag(`challenge-${id}-submissions`);
```
Comment (Hacksore) [inline apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts:41]: AFAIK you should only have to do this to get the data to update after deleting.

https://nextjs.org/docs/app/api-reference/functions/revalidateTag

apps/web/src/app/[locale]/challenge/[id]/solutions/_components/delete/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+    async function handleDeleteSolution() {
+        
+        try {
+            await deleteSolution(solution.id);
+            toast({
+                title: 'Solution Deleted',
+                variant: 'success',
+                description: 'The solution was successfully deleted.',
+            });
+            router.back();
```
Comment (Hacksore) [inline apps/web/src/app/[locale]/challenge/[id]/solutions/_components/delete/index.tsx:37]: Hmmm why do we call this?

Comment (zaCKoZAck0) [reply apps/web/src/app/[locale]/challenge/[id]/solutions/_components/delete/index.tsx:37]: because the route doesn't change even if the solution is deleted, so it results in this 
image

apps/web/src/utils/auth-guards.ts
```diff
@@ -4,3 +4,7 @@ import { RoleTypes } from '@repo/db/types';
 export function isAdminOrModerator(session: Session | null) {
   return [RoleTypes.ADMIN, RoleTypes.MODERATOR].some((i) => session?.user.role.includes(i));
 }
+
+export function isAuthor(session: Session | null, userId?: string | null) {
```
bautistaaa, line 8: the return type here should be boolean. can you wrap in `Boolean`

apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts
Comment (zaCKoZAck0) [reply apps/web/src/app/[locale]/challenge/[id]/solutions/_components/_actions.ts:41]: Ya, It doesn't have any effect here either, I've to try something to refetch/revalidate data after deletion.

apps/web/src/utils/auth-guards.ts
zaCKoZAck0: Sure

bautistaaa:
can you run `pnpm checks:fix`

Hacksore (approved):

PR #820, opened 2023-10-09
feat: adds beta token for early access

### LINKED ISSUE #807: [Feature]: enable closed beta feature
### Description

we want to only allow those on the waitlist to be able to sign up.

- [x] create a unique code per email that can be claimed during sign up (we're appending token column to csv import)
- [x] create a login page that will accept use token as a query param
- [x] figure out a way to stop account creation if an invalid token is provided
- [x] create a table to hold the tokens
- [x] import tokens into mailchimp (see first checkbox)

### Explanation

_No response_

bautistaaa:
## Description
Going for the simplest approach here
Overall changes

- adds a login page that will take a token as a param
  - validate token to see if it exists and hasnt been claimed(tokens are not 1:1 to emails)
- guard all public routes, IF early access is enabled redirect to `waitlist`
- you can ignore the scripts, its to facilitate manual work with mailchimp
- unsure: i removed the middleware code to redirect to waitlist and it now exists in every route. i think i tried to do it in the middleware but it didnt work to check session, ff, etc

## Related Issue

Closes #807


Hacksore:
i caused the mergies with server side nav, sorry not sorry.

apps/web/src/app/[locale]/login/page.tsx
```diff
@@ -0,0 +1,56 @@
...
+              By clicking continue, you agree to our{' '}
+              <Link href="/tos" className="hover:text-primary underline underline-offset-4">
+                Terms of Service
+              </Link>{' '}
+              and{' '}
+              <Link href="/privacy" className="hover:text-primary underline underline-offset-4">
+                Privacy Policy
+              </Link>
+              .
+            </p>
```
Comment (ayrock-dev) [inline apps/web/src/app/[locale]/login/page.tsx:45]: Good move 👍

packages/db/src/export-tokens.ts
```diff
@@ -0,0 +1,24 @@
+// @ts-nocheck
```
ayrock-dev, line 1: thanks for committing this for now, Epic Stack is doing something similar keeping loosey-goosey scripts in an `/other` dir

https://github.com/epicweb-dev/epic-stack/tree/main/other

apps/web/src/app/[locale]/explore/_components/index.tsx
```diff
@@ -3,11 +3,22 @@ import { ExploreSection } from './explore-section';
...
 export const dynamic = 'force-dynamic';
 
 export async function Explore() {
+  // early acces you must be authorized
+  const session = await getServerAuthSession();
+  const flags = await getAllFlags();
+
+  if (!session && flags.enableEarlyAccess) {
+    return redirect('/waitlist');
+  }
```
Comment (ayrock-dev) [inline apps/web/src/app/[locale]/explore/_components/index.tsx:20]: LG, these guards do the trick for now 🚀

packages/db/src/export-tokens.ts
bautistaaa: ah i should move it to a tmp folder!

apps/web/src/app/[locale]/explore/_components/index.tsx
Comment (bautistaaa) [reply apps/web/src/app/[locale]/explore/_components/index.tsx:20]: yeah its "good enough" imo cause its all temp and im lazy haha

apps/web/package.json
```diff
@@ -11,7 +11,7 @@
     "format:fix": "prettier . --write --cache --cache-location=\"node_modules/.cache/prettiercache\" --log-level=warn",
     "lint": "eslint . --cache --cache-location \"node_modules/.cache/.eslintcache\" --max-warnings 0",
     "lint:fix": "eslint . --fix --cache --cache-location \"node_modules/.cache/.eslintcache\" --max-warnings 0",
-    "start": "next start",
+    "start": "dotenv -e ../../.env next start",
```
QuiiBz, line 14: That's basically an alias to the `dev` script?

apps/web/src/app/[locale]/tracks/[slug]/page.tsx
```diff
@@ -10,7 +13,13 @@ interface Props {
...
 
 // todo: write a suspense skeleton...
-export default function Page({ params }: Props) {
+export default async function Page({ params }: Props) {
+  const session = await getServerAuthSession();
+  const flags = await getAllFlags();
+
+  if (!session && flags.enableEarlyAccess) {
+    return redirect('/waitlist');
+  }
```
Comment (QuiiBz) [inline apps/web/src/app/[locale]/tracks/[slug]/page.tsx:22]: We could extract this logic in a hook instead of duplicating it in many places, though I think the `redirect` will have to be duplicated

QuiiBz (changes requested):

apps/web/package.json
bautistaaa: if i run build i cant run start cause it wont pickup the env vars

bautistaaa: dev is specifically for dev (not prod build)

apps/web/src/app/[locale]/tracks/[slug]/page.tsx
Comment (bautistaaa) [reply apps/web/src/app/[locale]/tracks/[slug]/page.tsx:22]: eh actually ended up being kinda annoying to do, gonna ship it and strip it later

apps/web/package.json
QuiiBz: oh yeah I'm so dumb

QuiiBz (approved):

PR #823, opened 2023-10-10
fix: make challenge OG description renderable and deterministic

### LINKED ISSUE #819: [Bug]: Challenge OG images cant render markdown
### What happened?

Currently we use the description on the og image cards. The issue here is they are markdown and the card cannot render these. We should probably just replace with some generic text. I'd use chatgpt this but maybe one sentence about how its a challenge and if you can solve it or something corny... but not too corny

image

_No response_

### Relevant log output

_No response_

amorriscode:
## Description

- [x] Removes the markdown from OG descriptions by making it deterministic
- [x] Lints some files

## Related Issue

Closes #819

## Motivation and Context

Markdown isn't easily rendered in the OG image. I was originally going to take the first sentence of a challenge but this posed a few issues (eg. what if someone doesn't write in full sentences? what if markdown is in the first sentence? etc).

image

### DIFF
--- a/apps/og-image/app/api/challenge/route.tsx
+++ b/apps/og-image/app/api/challenge/route.tsx
@@ -81,7 +81,7 @@ export async function GET(req: Request) {
               tw="text-7xl font-bold truncate pr-48 -mt-2 overflow-ellipsis"
               style={{ fontWeight: '800 !important' }}
             >
-              {props.title === null ? null : props.title.split(" | ")[0]}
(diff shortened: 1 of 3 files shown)


Hacksore:
You gotta love the testing og in iMessage 😅

FWIW you can use http://localhost:4200 to get a preview app for this but maybe you just wanted to show what it'd look like in a real app.
image

PR #835, opened 2023-10-11
Add tracker for submit

Hacksore:
## Description

Please suggestion more events to add before launch.

## Related Issue

Closes #


packages/monaco/package.json
```diff
@@ -52,6 +52,7 @@
     "@monaco-editor/loader": "^1.3.3",
     "@monaco-editor/react": "^4.5.1",
     "@repo/ui": "workspace:^",
+    "@vercel/analytics": "^1.0.2",
```
Hacksore, line 55: a bit sus I have to add this here since it's in a separate package.

bautistaaa: you can pass the tracker through from the web package where we consume the component. i think the editor takes callbacks we define for both tests and user code. can just put the tracker i believe

odicho:
Here are some suggestions:
* Geolocation (interesting data)
* Device and browser data (will help with debugging)
* Advanced: Are we able to track time spent on a problem?
* Page views for each challenge (we can quickly figure out which ones are popular)
* Track the result as well: success/failure (will help gauge difficulty)

Slightly unrelated, but if we have a "Share" button to different platforms, we should implement utm_source for tracking traffic sources.

packages/monaco/package.json
Hacksore: for now I just only run `track` in the user face panel.

packages/monaco/src/index.tsx
```diff
@@ -27,6 +28,7 @@ export interface CodePanelProps {
   settingsElement: React.ReactNode;
   updatePlaygroundTestsLocalStorage?: (code: string) => void;
   updatePlaygroundCodeLocalStorage?: (code: string) => void;
+  track?: (event: string, props?: Record<string, AllowedPropertyValues>) => void;
```
bautistaaa, line 31: we cant pull this type from the package?

bautistaaa: nevermind youd have to install the package to get the types zzzzzzzz

bautistaaa (approved):

ayrock-dev:
lets goooo thanks @Hacksore!

PR #838, opened 2023-10-11
fix: fix error on deep link to deleted comment

ayrock-dev:
### Issue

visiting a deep link to a deleted comment renders the challenge page with the comment view open, but stuck in an "broken loading state"

i sus'd an error message in the server output:
```bash
web:dev: Invalid `prisma.comment.findMany()` invocation:
web:dev: 
web:dev: 
web:dev: Error in query graph construction: AssertionError("Invalid value for skip argument: Value can only be positive, found: -10")
web:dev:  X PrismaClientUnknownRequestError: 
web:dev: Invalid `prisma.comment.findMany()` invocation:
web:dev: 
web:dev: 
web:dev: Error in query graph construction: AssertionError("Invalid value for skip argument: Value can only be positive, found: -10")
web:dev:     at async getPaginatedComments (./src/app/[locale]/challenge/_components/comments/getCommentRouteData.ts:125:22)
```

ah-ha, looks like a `prisma.comment.findMany()` with a `{ skip: -10 }`

turns out that `skip` is calculated with an `index * 10`. `index` is calculated with a `comments.findIndex(. . .)`, and when the comment isn't found then `index` is `-1`. 

### Solution

stupid fix, but when we can't find the deep-linked comment id we can just not return pre-selected comment metadata. The client components already handle this case starting at page 0 (first page) with 0 skip.

### Testing Steps

* comment on a challenge
* click share on the comment to copy the link to clipboard
* delete the comment
* try to visit that link on your clipboard

After this fix, things successfully load.

### DIFF
--- a/apps/web/tests/e2e/authenticated/comment.test.ts
+++ b/apps/web/tests/e2e/authenticated/comment.test.ts
@@ -101,7 +101,7 @@ test.describe('create, edit, and delete comments', () => {
   });
 });
 
-test.describe('share comment', () => {
+test.describe('sharing comments', () => {
   const commentNumber = Math.floor(Math.random() * 1000);
   const parentComment = `Here is my comment ${commentNumber}`;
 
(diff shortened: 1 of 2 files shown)


bautistaaa (approved):
the goated e2e test

apps/web/tests/e2e/authenticated/comment.test.ts
```diff
@@ -127,4 +127,33 @@ test.describe('share comment', () => {
     await page.goto(`${page.url()}/comments/${comment.id}`);
     await expect(page.locator('div[id^=comment]', { hasText: parentComment })).toBeVisible();
   });
+
+  test('visiting a link to a deleted comment successfully loads page with remaining comments', async ({
+    page,
+  }) => {
+    const commentA = await prisma.comment.create({
```
Hacksore, line 134: my only comment should be using the browser/dom to create/delete records 😅

PR #844, opened 2023-10-12
Some Starting Challenges

dimitropoulos:
_This is a Draft PR to collect some work on starting challenges.  If you don't know why it's here, feel free to ignore._

My goal with this PR was to identify and complete all the drafts for challenges that one would need to take to be ready to do `Pick` without any knowledge gaps.  As you can see: there are quite a lot of dependencies of dependencies to get there.  The important point here with this graph is that it needs to be _possible_ to do things in some kind of order without presenting concepts that haven't been covered yet.  In other words, the knowledge graph has to be a directed acyclic graph.

Screenshot_20231014_134707

So, in this case, a track that assumes literally no TypeScript knowledge might go like:

1. primitive-types
1. type-aliases
1. type-unions
1. literal-types
1. typeof
1. keyof
1. index-signatures
1. indexed-types
1. mapped-object-types
1. generic-type-arguments
1. generic-constraints
1. pick


challenges/type-aliases/prompt.md
```diff
@@ -0,0 +1,88 @@
+## What Problem Do Type Aliases Solve?
+
+![mars_climate_orbiter](https://www.simscale.com/wp-content/uploads/2017/12/Mars_Climate_Orbiter_768.jpg)
+
+Did you know that a 125 _million_ dollar robotic space probe, the Mars Climate Orbiter, crashed into Mars because of a simple units conversion.
+
+Let's envision that code in TypeScript and see how type aliases can help.
```
bautistaaa, line 7: amazing

challenges/type-aliases/prompt.md
```diff
@@ -0,0 +1,88 @@
...
+  displayed: boolean;
+  selectable: boolean;
+  label: string;
+  expanded: boolean;
+}
+
+export type Rows = Row[];
+```
+
+That might seem nice because you can just import `Rows` and use it somewhere in your app where you have an array of Rows.  But avoid doing this.  You can just import `Row` and specify `Row[]` when you use it.  This is beneficial because it's simpler to read and immediately understand but also because it gives you access to the `Row` type, which you're probably going to want to work with _anyway_.
```
bautistaaa, line 82: the tip seems a little out of place. one concern is the line `it gives you access to the Row type` i feel people are going to want to know why which is distracting from the next line of solving a challenge

i dont feel strongly about this though

challenges/type-aliases/tests.ts
```diff
@@ -0,0 +1,41 @@
...
+  yearBuilt: Year;
+  crewCapacity: Count;
+  launchDate: Date;
+  isOperational: IsOperational;
+  propulsionSystem: string[];
+  payload: Payload[];
+}
+
+type test_Name = Expect<Equal<Name, string>>;
+type test_Year = Expect<Equal<Year, number>>;
```
bautistaaa, line 14: this challenge requires them to look at these test types which could be overwhelming for a beginner maybe?

challenges/type-unions/prompt.md
```diff
@@ -0,0 +1,70 @@
+## What Problem Type Unions Solve
+
+Type In a previous example for `type-aliases` we saw a problem.  We were writing code for a space shuttle and we wanted to make sure that we can't accidentally use the wrong units type.
```
bautistaaa, line 3:
```suggestion
In a previous example for `type-aliases` we saw a problem.  We were writing code for a space shuttle and we wanted to make sure that we can't accidentally use the wrong units type.
```

typo?

bautistaaa: not sure if we should say in a previous problem but maybe it makes sense in the context of a track but this can be completed out of order and ad hoc

challenges/type-unions/prompt.md
```diff
@@ -0,0 +1,70 @@
...
+```ts
+type Meters = number;
+type Miles = number;
+
+const landSpacecraft = (distance: Meters) {
+  // ... do fancy math ...
+}
+```
+
+But we noticed when we call `landSpacecraft` with the wrong literal type (see [the challenge on literal types](todo-link)), there's no error like we might expect:
```
bautistaaa, line 14: note: maybe we should use slugs instead of id so we can link easier

cc @Hacksore

challenges/type-unions/prompt.md
```diff
@@ -0,0 +1,70 @@
...
+
+That's a good start, but there's still no way for our function to tell the difference between the two, so we _spruce up_ the first two type aliases a bit.
+
+```ts
+type Meters = {
+  unit: 'meters';
+  value: number;
+};
+
+type Miles = {
```
bautistaaa, line 46: should we even go simpler and have discriminated unions as a separate challenge?

bautistaaa (commented):
AMAZING WORK!!!! 🚀

challenges/type-unions/prompt.md
dimitropoulos: for SEO purposes you should strongly strongly strongly consider using slugs.  my intention was that these ids are also slugs.

also, number IDs like the type challenges incorrectly conveys the sense that there's some kind of ordering or something (which, there isn't.  it's a graph of challenges not an ordered list).

bautistaaa: will make it happen today

challenges/generic-type-arguments/prompt.md
```diff
@@ -0,0 +1,95 @@
...
+
+```ts
+interface Row<T> {
+  label: string;
+  value: T;
+  disabled: boolean;
+}
+```
+
+> Note: It's common (and acceptable) in situations like this where we _really don't know anything else_ about the Generic type parameter to use a single letter like `T`.  But the moment you have multiple arguments (or more context for what this type will be) it's a good idea to use more descriptive names.
```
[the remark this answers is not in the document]
dimitropoulos: that's good feedback, thanks!

challenges/generic-type-arguments/prompt.md
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+> Note: you might notice that we're missing types for our type arguments!  Right now there's nothing stopping us from sending in wrong values like `GroceryItem<number[], boolean, { over: 9000 }>`.  We'll cover that in a future challenge on [`generic-constraints`](todo-link)
+
+## Solving This Challenge
+
+Similar to what we did with `AvocadoToast`, create a `CapreseSalad` type, and fill in the right values to make the tests pass.
+
+Then, create a new generic type for `GroceryStore`.  Let the tests guide you regarding what the final type should look like and what the inputs should be.
```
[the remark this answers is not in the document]
dimitropoulos: :)  Felt appropriate to give people _some_ kind of something challenging to do

challenges/keyof/prompt.md
```diff
@@ -0,0 +1,61 @@
...
+
+But what if you wanted to extract all these keys as a type?  You _certainly_ would not want to retype all those artist names.
+
+You can use `keyof`!
+
+## How To Use `keyof`
+
+`keyof` is special TypeScript syntax that you use before any type.
+
+In our case we _don't have_ a type, so we create one with the `typeof` operator:
```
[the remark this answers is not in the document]
dimitropoulos: yes absolutely.  added in 9cd64cd41e925c54b71bc1b5deed16cb1c1ab324

challenges/literal-types/prompt.md
```diff
@@ -0,0 +1,104 @@
...
+interface Ingredient {
+  name: Name;
+  inStock: InStock;
+  kilograms: Kilograms;
+}
+```
+
+But there's a problem.  Actually there are lots of problems that literal types solve but let's just look at one of them.
+
+Let's say you have a function like this:
```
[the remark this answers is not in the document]
dimitropoulos: I added some context in 3d07288a7fa462440ab85fcac63f19397b4591ed

there's a bit of a chicken-and-egg issue with sequencing this one and unions

challenges/literal-types/prompt.md
```diff
@@ -0,0 +1,104 @@
...
+    return true;
+  }
+  return false;
+}
+```
+
+## Using Literal Types
+
+```ts
+type Name = 'chocolate chips';
```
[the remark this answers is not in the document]
dimitropoulos: I've gone back and forth on that.  If you thought it was strange enough to even mention it then that's all the evidence that I need that I should go back to type instead of interface.  I did so in 3d07288a7fa462440ab85fcac63f19397b4591ed

challenges/generic-type-arguments/prompt.md
```diff
@@ -0,0 +1,95 @@
+## Why We Need Generic Type Arguments
+
+If primitive (and literal) types are data, you can think of generics as functions that operate on that data.
+
+And just like functions often need arguments, we need some way to provide inputs to generic types.
+
+## How To Use Generic Arguments
+
+Instead of using parenthesis to pass arguments, with generics we use angled brackets: `<` and `>`.
```
QuiiBz, line 9: Should we mention that it's *kinda* similar to generics in other languages (Java, C#, Rust...)?

challenges/generic-type-arguments/prompt.md
```diff
@@ -0,0 +1,95 @@
...
+
+```ts
+type AvocadoToast = {
+  name: 'Avocado Toast',
+  price: 12.99,
+  inStock: true
+}
+```
+
+> Note: you might notice that we're missing types for our type arguments!  Right now there's nothing stopping us from sending in wrong values like `GroceryItem<number[], boolean, { over: 9000 }>`.  We'll cover that in a future challenge on [`generic-constraints`](todo-link)
```
QuiiBz, line 89:
```suggestion
> Note: you might notice that we're missing types for our type arguments! Right now there's nothing stopping us from sending in wrong values like `GroceryItem<number[], boolean, { over: 9000 }>`.  We'll cover that in a future challenge on [`generic-constraints`](todo-link).
```

challenges/generic-type-arguments/prompt.md
QuiiBz, line 95:
```suggestion
Then, create a new generic type for `GroceryStore`. Let the tests guide you regarding what the final type should look like and what the inputs should be.
```

challenges/keyof/prompt.md
```diff
@@ -0,0 +1,61 @@
...
+  'Aerosmith': 4,
+  'R.E.M.': 4,
+  'Blur': 3,
+  'The Smashing Pumpkins': 5,
+  'Britney Spears': 3,
+  'Whitney Houston': 3,
+};
+```
+
+But what if you wanted to extract all these keys as a type?  You _certainly_ would not want to retype all those artist names.
```
QuiiBz, line 31: To make it even clearer, we could even add an example of manually writing an `Artists` union type, specifically showing that we're duplicating the keys, and then showing the solution with `keyof` that avoids this duplication.

challenges/keyof/prompt.md
```diff
@@ -0,0 +1,61 @@
...
+
+```ts
+type Artists = keyof CasettesByArtist;
+```
+
+## What `keyof` _returns_
+
+The result of `keyof` is always a union.
+
+> Even if there are 0 or 1 objects, it doesn't hurt to think of the result as a union.  If there are no elements in the result will be `never` (which is the empty set in "Set theory" terminology).  When there's only one key than the resulting type will be a single primitive or literal type.
```
QuiiBz, line 55:
```suggestion
> Even if there are 0 or 1 objects, it doesn't hurt to think of the result as a union. If there are no elements in the result will be `never` (which is the empty set in "Set theory" terminology).  When there's only one key than the resulting type will be a single primitive or literal type.
```

challenges/literal-types/prompt.md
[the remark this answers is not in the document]
dimitropoulos: update: I'm killing this whole thing and rewriting it.  it sucked.

challenges/literal-types/prompt.md
```diff
@@ -0,0 +1,106 @@
...
+type Kilograms = number;
+
+type Ingredient = {
+  name: Name;
+  inStock: InStock;
+  kilograms: Kilograms;
+}
+```
+
+But there's a problem.  Actually there are lots of problems that literal types solve but let's just look at one of them.
```
QuiiBz, line 65: Sorry for all these extra-space comments, I'll stop here but there is probably more of them 😅

```suggestion
But there's a problem. Actually there are lots of problems that literal types solve but let's just look at one of them.
```

challenges/pick/prompt.md
```diff
@@ -0,0 +1,149 @@
...
+}
+```
+
+That's a lot of properties!  But what if you have a function that only needs two properties from this type:
+
+```ts
+const pickYourPokemon = (pokemon: Pokemon) => {
+  const { type, name } = pokemon;
+
+  return `You picked the ${type}-type Pokemon ${name}!
```
QuiiBz, line 26:
```suggestion
  return `You picked the ${type}-type Pokemon ${name}!`
```

QuiiBz (commented):
Some quick suggestion, I'll go over everything this weekend. Amazing job!

challenges/primitive-types/prompt.md
```diff
@@ -0,0 +1,54 @@
...
+- `unknown`
+- `any`
+
+And while we're in the weeds for a moment, you should be aware that _capitalized_ variants of these things do actually exist: `Number`, `String`, `Boolean`, `Symbol`, `Object`.  But these refer to something quite different (i.e. JavaScript global objects) and should virtually never be used as types.
+
+## Solving This Challenge
+
+To solve this challenge: add primitive types to the code until the errors go away.
+
+You might need to add type annotations to variables (declared with `const` or types of arguments).  You might also need to add missing properties to objects.
```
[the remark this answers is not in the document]
dimitropoulos: I agree, but do you have any other ideas?  just not initializing them?

challenges/primitive-types/prompt.md
```diff
@@ -0,0 +1,54 @@
+## What Problem Primitive Types Solves
+
+Because the ability to create primitive types are the foundational idea behind TypeScript, asking asking what problem primitive types solve is more like asking "what problem does _TypeScript_ solve".
```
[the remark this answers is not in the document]
dimitropoulos: fixed in 96e215ef1d23db28575742ddb9ad14f0c4c2b1f5

challenges/primitive-types/prompt.md
```diff
@@ -0,0 +1,54 @@
...
+
+Think about it for a second: spellcheckers work because there are encoded with rules about the vocabulary and syntax of English.  Similarly, TypeScript has advanced knowledge about the vocabulary and syntax of JavaScript.  But TypeScript can do even better because you can specify the type of your data.  That allows TypeScript to make even better judgement calls about whether something you're trying to do is likely to be a bug (or not).
+
+## What Does "Primitive" mean in this context?
+
+And speaking of words and language, the usage of "primitive" in this sense is a bit of technical jargon.  If you look up "primitive" in the dictionary you'll find that the top definition is:
+
+> 1. relating to, denoting, or preserving the character of an early stage in the evolutionary or historical development of something.
+
+So basically, something it means "ancient" or "undeveloped". (??)
```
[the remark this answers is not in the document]
dimitropoulos: fixed in 96e215ef1d23db28575742ddb9ad14f0c4c2b1f5

challenges/primitive-types/prompt.md
```diff
@@ -0,0 +1,54 @@
+## What Problem Primitive Types Solves
+
+Because the ability to create primitive types are the foundational idea behind TypeScript, asking asking what problem primitive types solve is more like asking "what problem does _TypeScript_ solve".
+
+TypeScript is like a really really smart spellchecker.  It helps you work faster when you're at your best, but it also saves you from misspelling words you're under stress at 3am the night before a big term paper is due.
```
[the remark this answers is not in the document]
dimitropoulos: fixed in 96e215ef1d23db28575742ddb9ad14f0c4c2b1f5

challenges/type-aliases/prompt.md
dimitropoulos: when I've taught TypeScript to people in the past _from scratch_ I noticed that this is a common trap people fall into.  They get excited about aliases and then overuse it for everything to a point that makes the code harder to work with because you get a `Rows` type but you just want the `Row` to work with.

I reworked it in 13ea5e902571f7ec714f8fb6a6f6d20d897ca15f to hopefully read a little easier.  Is this better?

challenges/type-aliases/tests.ts
dimitropoulos: great great feedback: I made it _much_ more hand-hold-y 71a8d2998cd1371dc552423f0bf9724482a3152a

whadday think?

challenges/type-unions/prompt.md
dimitropoulos: conceptually speaking you can't complete this one, the way it's written, without having first learned of type aliases.  

concretely, the reason I did a short little recap is precisely so that you don't _really_ have to actually take the `type-aliases` one first.  my thinking is that the reality is for these first few that are REALLY foundational, the likelihood is that there is a little bit of an in-built order (only a little though! and only because of what's natural in terms of building up basic concepts!)

(fixed the typo in 7b97897c320a5bd616e2b579053e4c90621e62be)

challenges/type-unions/prompt.md
dimitropoulos: I'd love that (and, indeed, actually started that way) but the problem I had was demonstrating _why it's useful_ without some kind of concrete example.  I guess that doesn't have to happen in _every single challenge_ but it's a goal of mine to try to keep them practical.

any ideas on other examples to use?

challenges/generic-type-arguments/prompt.md
dimitropoulos: ehh.  personally I don't know what value it adds.  there are enough that don't anyway (Go, Carbon, Haskell, etc.).  but I don't feel strongly at all about not adding it.  You have push access on my fork, please (really, seriously) feel absolutely totally free to push this if you'd like.

challenges/generic-type-arguments/prompt.md
dimitropoulos: 79e61006ed919ad42f1e91e016f6871d8aabc15a

challenges/literal-types/prompt.md
dimitropoulos: see https://github.com/typehero/typehero/pull/844#discussion_r1358504467

challenges/generic-type-arguments/prompt.md
dimitropoulos: see https://github.com/typehero/typehero/pull/844#discussion_r1358504467

challenges/pick/prompt.md
dimitropoulos: good catch: fixed in 891ee1ba12f9584965daa2b3e0d407b22847e81b

challenges/keyof/prompt.md
dimitropoulos: I gave it a try: I'm not sure what I think but it seems ok?  what do you think:

932440f34d6cdcf903eb199ec8e64da8e3802856

challenges/type-unions/tests.ts
```diff
@@ -0,0 +1,117 @@
+import { Expect } from 'type-testing';
+import { expect } from 'jest';
```
Hacksore, line 2: @dimitropoulos @ggrandi @bautistaaa is this technically possible with our current implementation?

dimitropoulos: glad you asked! I don't think it is yet, but to me it's among a short list of things that are absolutely critical.  It doesn't even have to be jest or whatever.  but being able to run actual javascript (and test it) is one of the biggest problems with type-challenges because it locks the entire project into the type realm _only_.  there are lots of important grounds to cover (i.e. practical things like we all do in our daily life with TypeScript) that you simply cannot do with types alone.  e.g. a challenge I'm planning on how to make generic constraints for functions (a pretty important skill) really doesn't work unless you can write a function.

as for the particular implementation here, again, doesn't matter.  I'd guess that we'll need to do some kind of `describe` block first or something if we want to accomplish actually running the tests.

dimitropoulos: oh, another thought, I haven't looked into this particular space of the testing world in a bit (assertion libraries) but https://github.com/power-assert-js/power-assert used to be a really fantastic choice (and maybe it still is).  it has no API, which makes it really easy to use.  You just construct your assertions so that they return booleans.

bautistaaa: we’re going to need some time to research the js capability 

we’d need a console to output logs maybe? but mainly sounds like you just need those tests to pass

this potentially might take some time. is it possible to write a challenge within these constraints for now 

ill prioritize this research this week

dimitropoulos: > is it possible to write a challenge within these constraints for now

In most cases, yes.  We can just comment out this one's usage of `expect` for now

PR #851, opened 2023-10-12
Use zod schema on server actions

Hacksore:
## Related Issue

Closes #


apps/web/src/app/[locale]/settings/_components/settings.action.ts
```diff
@@ -3,26 +3,36 @@
...
+export async function updateProfile(profileData: ProfileSchema) {
   const session = await getServerAuthSession();
 
   // 1. Checks that the user is logged in
-  if (!session?.user.id) return 'unauthorized';
+  if (!session?.user.id) return { error: 'unauthorized' };
 
-  // 2. Update the user bio field in the db
+  // 2. test schema validation with zod
+  try {
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/settings/_components/settings.action.ts:20]: can we just throw so we can try catch in consumer. not a fan of having to check error cause it just ends like a fetch we have at home. we also dont need to display the real error to users in most cases so returning isnt needed. in cases that it is we can handle that ad hoc

thoughts @QuiiBz

PR #866, opened 2023-10-14
fix: landing page metadata

### LINKED ISSUE #859: [Bug]: Landing page has `<head>` metadata from other page routes
### What happened?

During build, metadata from the landing page is being generated using information from other routes, eventually jumping around between multiple children routes on different build results. This seems to be a bug on Next's end.

One piece of context I'll put here is that the "jumping around" nature of the page it chooses in order to generate the landing page metadata seems to also include the landing page itself, which led me to believe that the problem had been previously solved by some of the first commits. Keep that in mind when debugging locally. 

Also important, the bug is not reproductible in development mode. You'll need to test changes by running and serving a build in order to see the bug occurring.

References to issues that could be related found by @QuiiBz: 
- https://github.com/vercel/next.js/issues/49988

All

### Relevant log output

_No response_

ernestoresende:
## Description
During build, metadata from the landing page is being generated using information from other routes, eventually jumping around between multiple children routes on different build results. This seems to be a bug on Next's end.

One piece of context I'll put here is that the "jumping around" nature of the page it chooses in order to generate the landing page metadata seems to also include the landing page itself, which led me to believe that the problem had been previously solved by some of the first commits. Keep that in mind when debugging locally. 

Also important, the bug is not reproductible in development mode. You'll need to test changes by running and serving a build in order to see the bug occurring.

References to issues that could be related found by @QuiiBz: 
- https://github.com/vercel/next.js/issues/49988

## Related Issue

Closes #859 

## Motivation and Context

- #859

### DIFF
--- a/apps/web/src/app/[locale]/layout.tsx
+++ b/apps/web/src/app/[locale]/layout.tsx
@@ -3,7 +3,6 @@ import { Analytics } from '@vercel/analytics/react';
 import '../../styles/globals.css';
 
 import { Inter } from 'next/font/google';
-import type { Metadata } from 'next';
 import { Toaster } from '@repo/ui/components/toaster';
 import { Providers } from './providers';
 import { buildMetaForDefault } from '../metadata';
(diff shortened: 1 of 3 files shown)


bautistaaa (approved):

QuiiBz (changes requested):
Just to prevent accidental merge, as it doesn't fix the issue :(

ernestoresende:
Yeah, as @QuiiBz said, this doesn't fix the issue, and at this point, I've ran out of ideas on how to solve it, and if it's even solvable with application-side code without delving in some workarounds.

I'll update the PR description and add some notes on the bug reproduction if anyone else wants to tackle this.

bautistaaa:
going to close

Issue #865, opened 2023-10-14
[Feature]: slugify tracks and challenges


bautistaaa:
### Description

read title

PR #870, opened 2023-10-15
chore: use slugs for tracks + challenges

bautistaaa:
## Description
i did this really fast so potentially some buggies and some things done wrong. will leave comments on the parts i think are icky

## Related Issue

Closes #https://github.com/typehero/typehero/issues/865


apps/web/src/app/[locale]/challenge/[slug]/solutions/[solutionId]/getSolutionIdRouteData.ts
```diff
@@ -4,11 +4,15 @@ import { type Session } from '@repo/auth/server';
 import { redirect } from 'next/navigation';
 
 export const getSolutionIdRouteData = cache(
-  async (challengeId: string, solutionId: string, session: Session | null) => {
+  async (slug: string, solutionId: string, session: Session | null) => {
+    const challenge = await prisma.challenge.findFirstOrThrow({
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/[slug]/solutions/[solutionId]/getSolutionIdRouteData.ts:8]: we used to be able to get id from the url but since its slug i need to query and get the id.

i could make slug a foreign key to solutions but unsure if thats a good idea.. id seems the only reasonable FK here

apps/web/src/app/[locale]/challenge/[slug]/submissions/[[...catchAll]]/page.tsx
```diff
@@ -40,9 +41,13 @@ export default async function SubmissionPage({ params: { id } }: Props) {
 }
 
 export type ChallengeSubmissions = NonNullable<Awaited<ReturnType<typeof getChallengeSubmissions>>>;
-const getChallengeSubmissions = cache((userId: string, challengeId: string) => {
+const getChallengeSubmissions = cache(async (userId: string, slug: string) => {
+  const challenge = await prisma.challenge.findFirstOrThrow({
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/[slug]/submissions/[[...catchAll]]/page.tsx:45]: we used to be able to get id from the url but since its slug i need to query and get the id.

i could make slug a foreign key to submissions but unsure if thats a good idea.. id seems the only reasonable FK here

PR #871, opened 2023-10-15
fix:admin tab loading

### LINKED ISSUE #810: [Bug]: Admin tabs do not load data until you refresh
### What happened?

Often times when you load admin portal the tabs are either stuck in a loading state or nothing appears until you refresh.

_No response_

### Relevant log output

_No response_

shlok-2003:
## Description
Removed some return statments

## Related Issue
Closes #810


bautistaaa:
i dont understand how this fixes the problem

shlok-2003:
Hey @bautistaaa, after these changes, the following issue is not being reproduced on my localhost. Is there any way to test it more further

apps/admin/src/app/_components/tracks/index.tsx
```diff
@@ -6,14 +6,20 @@ import { ActionBar } from './action-bar';
...
+
 export function ManageTracks() {
   const { data, isLoading } = useQuery(['admin-tracks'], () => getTracks());
-  if (isLoading || !data) return <div>Loading...</div>;
 
   return (
     <div>
-      <ActionBar />
-      <DataTable data={data} columns={columns} />
+      {!data || isLoading ? (
```
ja153903, line 15: This effectively doesn't change anything. It just moves the condition from its own if-statement down to the return

apps/admin/src/app/_components/challenge-reviews/index.tsx
```diff
@@ -8,11 +8,9 @@ import { columns } from './columns';
 export function ChallengeReviews() {
   const { data, isLoading } = useQuery(['challenge-reviews'], () => getPendingChallenges());
 
-  if (isLoading || !data) return null;
-
   return (
     <div className="w-full rounded-md border">
-      <DataTable data={data} columns={columns} />
+      {data ? <DataTable data={data} columns={columns} /> : ''}
```
ja153903, line 13: Similar to the other change, you're just displacing where the empty check is. This doesn't solve the supposed issue.

ja153903 (commented):
@bautistaaa I think we need a reliable way to reproduce this bug.

shlok-2003:
Ok, I'll see if I can find any errors in the api's

bautistaaa:
> @bautistaaa I think we need a reliable way to reproduce this bug.

if there was i would have written it. unfortunately its mainly on initial load. its not high priority as its admin only just annoying

shlok-2003:
Hey @ja153903, as of now, I am unable to find any issues in the codebase.

PR #878, opened 2023-10-16
831 : Fix UI bug in solution code editor 

### LINKED ISSUE #831: [Bug]: markdown editor bugged in dark mode
### What happened?

initially displaying the markdown editor shows incorrectly. its not always consistently broken though and toggling the light and dark mode fixes it. my initial hunch is on mount its not detecting the theme properly

CleanShot 2023-10-10 at 14 08 27

_No response_

### Relevant log output

_No response_

DevWig:
Update theme setting logic to include instance where theme is set to 'system' instead of 'dark' or 'light' to prevent using light theme variables in dark mode

Changed the conditional that sets the theme for the page in the solution-editor file

I chose to keep the same ternary operator style statement and just manipulate it so that it picks up all three possible cases, as well as sets a default of 'system' in case 'theme' is not defined.
However, in rich-markdown-editor there is a useEffect hook with a switch-case block that is handling this same thing for another editor. I can change what I did to match that if I need to, just let me know.

## Related Issue
Closes #831 

## Motivation and Context
This change fixes the bug that occurs when opening the solutions editor by adding/submitting a solution within a challenge where the color variables for the editor are for the light theme when 'theme' is set to 'system'

## How Has This Been Tested?
Log in
Make sure the current theme is dark mode
View a challenge card and enter a solution
Navigate to the 'Solutions' tab in the editor to the left
Click the '+ Solution' button and visually verify the colors are correct for dark mode


bautistaaa:
can you fix the conflict

DevWig:
> can you fix the conflict

Yessir! Just pushed it

bautistaaa (approved):

PR #881, opened 2023-10-17
contributors grid redesign, fixes #837

### LINKED ISSUE #837: [Feature]: Redesign the contributors layout
### Description

We have reach a point where we have too many contributors for the community section. We need to make the rows wider so it doesnt grow so tall or some other design. 

CleanShot 2023-10-10 at 22 44 23@2x

### Explanation

_No response_

PickleNik:
https://github.com/typehero/typehero/assets/31113245/be14e98b-143d-487b-bf6d-9943b0cbbce5

Screenshot 2023-10-16 at 11 34 48 PM
Screenshot 2023-10-16 at 11 35 09 PM


bautistaaa:
the current version i think is better(more impactful cause you see more) for desktop but this version is much better for mobile. im okay going with this though

PickleNik:
> the current version i think is better(more impactful cause you see more) for desktop but this version is much better for mobile. im okay going with this though

changed xl back to the current version btw, but with the new hover effect

bautistaaa (approved):

PR #890, opened 2023-10-17
feat: integrate ata

bautistaaa:
## Description
We want to be able to import 3rd party types in our challenges.

- [x] fetch types on change
- [x] fetch types on load
- [x] fetch types for test model
- [x] fetch types for user model
- [x] types successfully load in monaco
- [x] debounce the event for fetching types

### Sources
https://github.com/microsoft/TypeScript-Website/tree/9150ed65461c7ff0b3ec1b2085a30d26843fe3ec/packages/ata

## Related Issue

Closes https://github.com/typehero/typehero/issues/329

PR #906, opened 2023-10-19
change test types and ingest script

bautistaaa:
## Description
importing is proving to be difficult. WIP PR: https://github.com/typehero/typehero/pull/890

in the mean time we can just full the test types locally and remove the imports from the test panel while this is still being worked on.

## Related Issue

Closes #


bautistaaa:
cc [dimitropoulos](https://github.com/dimitropoulos)

PR #908, opened 2023-10-20
PoC server side testing

Hacksore:
It's kinda slow but there try this:
```sh
curl --location 'http://127.0.0.1:3005/api/test' \
--header 'Content-Type: application/json' \
--data '{
    "code": "const str: Foo = '\''ok'\''",
    "testCase": "type Foo = number"
}'
```

Output:
```
{
    "errors": [
        "index.ts (1,7): Type 'string' is not assignable to type 'number'."
    ]
}
```

## Description
I was going to use CF to do this but there are some serious limitations. We can't read FS on workers and we need to load all the libs/* from typescript to not get errors 😬.

So now this is a next API app 💀

## Related Issue
Closes nothing


Hacksore:
Forgor about this but there is some work being done in a different repo to make this "work" then we can think about integrating it into typehero.

If you wanna pull up to that repo it's here:
https://github.com/Hacksore/typecheck-service

The main goals are to have type-checking server side as well as perf testing.

cc @ernestoresende @ggrandi

PR #915, opened 2023-10-20
fix(tracks): prompt for user login before track enrollment

### LINKED ISSUE #912: [Bug]: Enrolling in a track when not logged in should prompt user to login
### What happened?

Currently a generic toast appears that is not very helpful

_No response_

### Relevant log output

_No response_

aakash19here:
Closes #912 

Screenshot 2023-10-20 at 4 06 29 PM

### DIFF
--- a/apps/web/src/app/[locale]/tracks/_components/track.action.ts
+++ b/apps/web/src/app/[locale]/tracks/_components/track.action.ts
@@ -14,7 +14,7 @@ import { createTrackGridCacheKey } from './track-grid';
 export async function enrollUserInTrack(id: number) {
   const session = await getServerAuthSession();
   if (!session) {
-    throw new Error('You must be logged in to enroll in a track.');
+    throw new Error('User is not logged in');
   }
 
   await prisma.track.update({
--- a/apps/web/src/app/[locale]/tracks/_components/track.action.ts
+++ b/apps/web/src/app/[locale]/tracks/_components/track.action.ts
@@ -16,19 +16,22 @@ export async function enrollUserInTrack(id: number) {
   if (!session) {
     throw new Error('User is not logged in');
   }
-
-  await prisma.track.update({
-    where: {
-      id,
-    },
-    data: {
-      enrolledUsers: {
-        connect: {
-          id: session?.user.id,
+  try {
+    await prisma.track.update({
+      where: {
+        id,
+      },
+      data: {
+        enrolledUsers: {
+          connect: {
+            id: session?.user.id,
+          },
         },
       },
-    },
-  });
+    });
+  } catch {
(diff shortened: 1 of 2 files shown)


bautistaaa (changes requested):
this assumes every error is cause of not being logged in

aakash19here:
> this assumes every error is cause of not being logged in

I had added ternery a ternery operator for showing the error in the previous commit. If that works I'll add conditional rendering for the title too

bautistaaa:
> > this assumes every error is cause of not being logged in
> 
> I had added a ternery operator for showing the error conditionaly in the previous commit. If that works I'll add conditional rendering for the title too

you should only display the login error if theres no session else show the original. you can either return 'unauthorized' from the action like we do in other actions or use the error(i dont think weve done this before)

apps/web/src/app/[locale]/tracks/_components/track.action.ts
```diff
@@ -16,19 +16,22 @@ export async function enrollUserInTrack(id: number) {
...
+          connect: {
+            id: session?.user.id,
+          },
         },
       },
-    },
-  });
+    });
+  } catch {
+    throw new Error('Internal server error');
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/tracks/_components/track.action.ts:33]: why do we need a catch

aakash19here:
Removed id from the WHERE clause to cause an internal server error
Screenshot 2023-10-20 at 11 09 59 PM


Logged out and enrolled in the track 
Screenshot 2023-10-20 at 11 12 06 PM

apps/web/src/app/[locale]/tracks/_components/track.action.ts
Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track.action.ts:33]: Cause there can be an error while fetching data from db , so it makes sense to add it in a try catch block and then throw the error

Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track.action.ts:33]: if there are any other good practices I will be happy to hear it out from you : )

Comment (bautistaaa) [reply apps/web/src/app/[locale]/tracks/_components/track.action.ts:33]: youre just catching and rethrowing, wont it work the same way without it

Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track.action.ts:33]: Wait , I understood what you are saying 
I will add a ternary to just check if user is unauthorized , else i will throw the the other error message , this will remove the try catch

Comment (aakash19here) [reply apps/web/src/app/[locale]/tracks/_components/track.action.ts]: > youre just catching and rethrowing, wont it work the same way without it

yes , thats what I thought later

aakash19here:
@bautistaaa I've made the requested changes :)

bautistaaa (approved):

PR #945, opened 2023-10-23
feat: Inherited font sizes for challenge prompts

### LINKED ISSUE #944: [Feature]: Challenge view cleanup
### Description

Currently the default of the editor font and prompt are different. Prompt (16px) and Editor (13px). They should align at least by default.

Additionally, the code blocks in the prompt should match the same font as the rest of the prompt

- [ ] default editor to 16px or whatever the size is in the markdown prompt
- [ ] code blocks in the prompt should be the same font size as the rest of the prompt
- [ ] decrease line height in the markdown prompt to 1.5rem

### Explanation

This will make the whole reading/writing experience feel more connected.

oreoluwa-bs:
## Description

Changed the default font size of prompts to 13px and made the font size in the code block inherited

## Related Issue

Closes #944 

## Motivation and Context

#944 

## How Has This Been Tested?

No

Screenshot 2023-10-23 at 23 56 00

### DIFF
--- a/packages/ui/src/components/markdown.tsx
+++ b/packages/ui/src/components/markdown.tsx
@@ -76,19 +76,6 @@ export function Markdown({ children, className }: { children: string; className?
               customStyle={{ fontSize: 'inherit' }}
               codeTagProps={{
                 style: {
-                  // Default styles
-                  fontFamily:
-                    'Menlo, Monaco, Consolas, "Andale Mono", "Ubuntu Mono", "Courier New", monospace',
-                  color: 'rgb(212, 212, 212)',
-                  textShadow: 'none',
-                  direction: 'ltr',
-                  textAlign: 'left',
-                  whiteSpace: 'pre',
-                  wordSpacing: 'normal',
-                  wordBreak: 'normal',
-                  tabSize: 4,
-                  hyphens: 'none',
-
                   // overrides
                   fontSize: 'inherit',
                   lineHeight: 'inherit',
--- a/packages/ui/src/components/markdown.tsx
+++ b/packages/ui/src/components/markdown.tsx
@@ -73,6 +73,27 @@ export function Markdown({ children, className }: { children: string; className?
               style={syntaxHighlighterTheme} // theme
               wrapLines
               wrapLongLines
+              customStyle={{ fontSize: 'inherit' }}
+              codeTagProps={{
+                style: {
+                  // Default styles
+                  fontFamily:
+                    'Menlo, Monaco, Consolas, "Andale Mono", "Ubuntu Mono", "Courier New", monospace',
+                  color: 'rgb(212, 212, 212)',
+                  textShadow: 'none',
+                  direction: 'ltr',
+                  textAlign: 'left',
+                  whiteSpace: 'pre',
+                  wordSpacing: 'normal',
+                  wordBreak: 'normal',
+                  tabSize: 4,
+                  hyphens: 'none',
+
+                  // overrides
+                  fontSize: 'inherit',
+                  lineHeight: 'inherit',
+                },
+              }}
               {...props}
             >
               {String(children).replace(/\n$/, '')}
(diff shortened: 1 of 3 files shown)


oreoluwa-bs:
@bautistaaa @odicho Hope this is good

maxdemaio:
Quick question, can you confirm that the changes you made to `packages/monaco/src/settings-store.ts` will override what is in the `apps/web/public/min/vs/editor` folder with the minified JS? I tried changing the `settings-store.ts` on my local and it doesn't rebuild so wondering if it worked for you.

bautistaaa:
no its not what the ticket is asking for

- the prompt default is 16px so the editor should default to 16px.
- you didnt make the line heigh adjustments 


review the checkboxes in the original ticket

maxdemaio:
> no its not what the ticket is asking for
> 
> * the prompt default is 16px so the editor should default to 16px.
> * you didnt make the line heigh adjustments
> 
> review the checkboxes in the original ticket

I think the change from `leading-8` to `leading-6` will add the line height changes to the markdown prompt from 2rem to 1.5rem! Agreed on everything else 👍🏻

oreoluwa-bs:
> Quick question, can you confirm that the changes you made to `packages/monaco/src/settings-store.ts` will override what is in the `apps/web/public/min/vs/editor` folder with the minified JS? I tried changing the `settings-store.ts` on my local and it doesn't rebuild so wondering if it worked for you.

I had to clear editor settings in local storage for the changes to update

oreoluwa-bs:
How about this
Screenshot 2023-10-24 at 01 45 08

packages/ui/src/components/markdown.tsx
```diff
@@ -73,6 +73,27 @@ export function Markdown({ children, className }: { children: string; className?
...
+                    'Menlo, Monaco, Consolas, "Andale Mono", "Ubuntu Mono", "Courier New", monospace',
+                  color: 'rgb(212, 212, 212)',
+                  textShadow: 'none',
+                  direction: 'ltr',
+                  textAlign: 'left',
+                  whiteSpace: 'pre',
+                  wordSpacing: 'normal',
+                  wordBreak: 'normal',
+                  tabSize: 4,
+                  hyphens: 'none',
```
odicho, line 90: Is there a specific reason for this?

odicho:
Just left one comment about the additional styling you added in the codeTagProps. 

Otherwise, it looks good to me:
1. The syntax highlighter in the description is 1rem (16px)
2. The code editor's default font size is 16px
3. Markdown line height is 1.5rem instead of 2rem.

packages/ui/src/components/markdown.tsx
oreoluwa-bs: Added it because the code blocks had an inline fontsize of 13px

odicho: If you don't include this I believe everything should be working as normal. I tried it locally without the added default styles in here.

oreoluwa-bs: Everything else except the font family remained the same. Although the fonts look similar
Screenshot 2023-10-24 at 09 31 29

odicho: I think that's fine though, isn't that what the font was before adding this? For example, if you remove the entire `customStyle` and `codeTagProps`, are those the fonts you see as well?

Just trying to make sure we don't have redundant default styles here if they're already included (except for fontSize and lineHeight)!

odicho: Otherwise, if the font actually changes to something else when you introduce the `customStyle` and `codeTagProps` then I think it's okay to leave the `fontFamily` as you have it and remove everything else!

odicho: Oki, then it looks like it is resetting to some other fonts without it. I think how you included it is good to go then. I think it looks good! I'll let Chris take a look in the morning and he can merge if he thinks it's okay as well!

PR #947, opened 2023-10-24
new(challenge): generic function arguments

dimitropoulos:
- added a new challenge for generic-function-arguments
- added a draft of a document on guidelines for creating challenges

### DIFF
--- a/challenges/generic-function-arguments/prompt.md
+++ b/challenges/generic-function-arguments/prompt.md
@@ -6,7 +6,7 @@ In many situations, we don't need functions to be generic.  The function might b
 
 ## Understanding Generic Syntax
 
-In a previous challenge on [generic type arguments](todo-generic-type-arguments) we saw a `Row` type that looked like this:
+In a previous challenge on [generic type arguments](https://typehero.dev/challenge/generic-type-arguments) we saw a `Row` type that looked like this:
 
 ```ts
 interface Row<T> {
--- a/challenges/generic-function-arguments/prompt.md
+++ b/challenges/generic-function-arguments/prompt.md
@@ -1,0 +1,80 @@
+## Why Functions Need Generics
+
+One of the most common needs a TypeScript developer faces is the need to pass generic type arguments to functions.  Then, those functions can use and manipulate those parameters.
+
+In many situations, we don't need functions to be generic.  The function might be written in such a way as to only work for one type of data (e.g. `string` or `number`).  But, other times, we write functions where we want to pass the data type along from input to output.
+
+## Understanding Generic Syntax
+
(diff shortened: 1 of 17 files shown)


challenges/generic-function-arguments/prompt.md
```diff
@@ -0,0 +1,80 @@
+## Why Functions Need Generics
+
+One of the most common needs a TypeScript developer faces is the need to pass generic type arguments to functions.  Then, those functions can use and manipulate those parameters.
+
+In many situations, we don't need functions to be generic.  The function might be written in such a way as to only work for one type of data (e.g. `string` or `number`).  But, other times, we write functions where we want to pass the data type along from input to output.
+
+## Understanding Generic Syntax
+
+In a previous challenge on [generic type arguments](todo-generic-type-arguments) we saw a `Row` type that looked like this:
```
bautistaaa, line 9: we can probably start doing `https://typehero.dev/challenge/my-slug`

challenges/generic-function-arguments/prompt.md
```diff
@@ -0,0 +1,80 @@
...
+// Thing 1: Start of JSX
+const createRow = <T>Some JSX stuff!</T>;
+
+// Thing 2: start of generic function
+const createRow = <T>(someArg: T) => {
+  /* some function stuff */
+}
+```
+
+Because of this ambiguity, we need to do something to _disambiguate_ these two situations to the TypeScript compiler.  In the early days it used to be common to do use `<T extends unknown>` (or `<T extends any>` in the days before `unknown` existed in TypeScript).  But over time people realized there's a shorter way: `<T,>`.  Because `,` is invalid in JSX, but valid in TypeScript to separate multiple arguments, it's sufficient to use this to disambiguate the two situations to the TypeScript compiler.
```
bautistaaa, line 64:
```suggestion
Because of this ambiguity, we need to do something to _disambiguate_ these two situations to the TypeScript compiler.  In the early days it used to be common to use `<T extends unknown>` (or `<T extends any>` in the days before `unknown` existed in TypeScript).  But over time people realized there's a shorter way: `<T,>`.  Because `,` is invalid in JSX, but valid in TypeScript to separate multiple arguments, it's sufficient to use this to disambiguate the two situations to the TypeScript compiler.
```

bautistaaa: theres a lot of double spaces after periods. not a big deal but just noticing it

bautistaaa (approved):
some small edits

challenges/generic-function-arguments/prompt.md
dimitropoulos: I fixed all slugs in https://github.com/typehero/typehero/pull/947/commits/f3942701e4005c43f4190d3ce4b706af0842b6b7

bautistaaa (approved):

PR #952, opened 2023-10-25
fix: bugged challenges

QuiiBz:
Fixing some challenges as I find issues


bautistaaa (approved):

ayrock-dev (approved):

challenges/literal-types/solutions/1.ts
```diff
@@ -11,8 +11,7 @@ type LiteralFunction = (a: number, b: number) => number;
 const literalString = 'Ziltoid the Omniscient';
 const literalTrue = true;
 const literalNumber = Math.random() > 0.5 ? 1 : 2;
-const almostPi = 3.14159265358979323846264338327950288419716939937510582097494459230781640628620899862803482534211706798214808651328230664709384460955058223172535940812848111745028410270193852110555964462294895493038196442881097566593344612847564823378678316527120190914564856692346034861045432664821339360726024914127372458700660631558817488152092096282925409171536436789259036001133053054882046652138414695194151160943305727036575959195309218611738193261179310511854807446237996274956735188575272489122793818301194912983367336244065664308602139494639522473719070217986094370277053921717629317675238467481846766940513200056812714526356082778577134275778960917363717872146844090122495343014654958537105079227968925892354201995611212902196086403441815981362977477130996051870721134999999837297804995105973173281609631859502445945534690830264252230825334468503526193118817101000313783875288658753320838142061717766914730359825349042875546873115956286388235378759375195778185778053217122680661300192787661119590921642019891337
```
dimitropoulos, line 14: this was here on purpose: to force someone to learn to copy/paste the literal type.  otherwise, for the `literalObject` section that comes next they might just write it by hand (which is error prone and also reinforces a bad habit when working with literal types).

What were your thoughts on removing it?

challenges/mapped-object-types/tests.ts
```diff
@@ -10,52 +12,53 @@ type MoviesByGenre = {
   thriller: 'The Silence of the Lambs';
 };
 
-type Example = MovieInfoByGenre<MoviesByGenre>;
-
-const example: Example = {
```
dimitropoulos, line 15: why remove this in favor of `Expect<Equal<`?

challenges/primitive-data-types/tests.ts
```diff
@@ -1,5 +1,5 @@
 type test_age = Expect<NotLiteral<number, typeof age>>;
-type test_name = Expect<NotLiteral<string, typeof name>>;
+type test_name = Expect<NotLiteral<string, typeof artistName>>;
```
dimitropoulos, line 2: if you're going to change the name then also change the name on the test

```suggestion
type test_artistName = Expect<NotLiteral<string, typeof artistName>>;
```

challenges/primitive-data-types/user.ts
```diff
@@ -24,6 +24,6 @@ const musicianInfo = ({
 };
 
 musicianInfo({
-  name,
+  name: artistName,
```
dimitropoulos, line 27: I don't really care one way or another, but sorta confused on this change.  if it was to avoid the shorthand syntax (which, is totally understandable: I went back-and-forth on this as well) then why not also change `age`?

challenges/literal-types/solutions/1.ts
bautistaaa: lets put it back. unsure if he removed it cause it was actually breaking tests

challenges/primitive-data-types/user.ts
bautistaaa: if you paste your challenge into ts playground it clashes with `name` from `lib.dom.ts`

challenges/mapped-object-types/tests.ts
bautistaaa: are these functionally equivalent?

bautistaaa (approved):

challenges/primitive-data-types/user.ts
dimitropoulos: ahhhhhh ok.  we can patch that out with the tsconfig but sure

challenges/literal-types/solutions/1.ts
dimitropoulos: it's no big deal: can be removed if it's a problem

challenges/mapped-object-types/tests.ts
dimitropoulos: well as far as I'm aware they're functionally equivalent, in which case we should revert to the original code (since it's much more "real world"

dimitropoulos: turns out this change broke the example (see https://github.com/typehero/typehero/issues/1019) and was fixed in https://github.com/typehero/typehero/pull/1009

PR #957, opened 2023-10-25
Update Settings/Profile with Small Changes

maxdemaio:
## Description

- Removes the links from the settings page because they're un-needed
- Add null check for user links length
- Equal alignment of profile items on left side

## Motivation and Context

Working on this with Trash when deploying to prod and we saw some visual issues

## How Has This Been Tested?

Visually w/ CSS and Browser

Settings (without links and equal gap with profile)

image

Profile (re-alignment with equal gap now):
image


PickleNik (approved):
less is more

PR #959, opened 2023-10-26
Fix Settings Page Link DB Call

maxdemaio:
## Description

Instead of `upsert`, I now am using a `deleteMany` and then creating up to 4 records.

In `UserLink`, the `user` field is a relation to the `User` model and we can't directly set it in a `createMany` operation since it's a nested operation. So to create multiple records with relations, I created an array of `userLinks` using the `create` method within a `prisma.$transaction` block.

## Motivation and Context

This fixes an issue where you can have more than 4 links. This is because we used to use `upsert`, but since it didn't have an `id` it couldn't update and only insert. This was causing an error on the `/settings` page because there was a negative amount of links trying to be shown because we do `4- n` but if `n > 4` it's negative.

## How Has This Been Tested?

Updating links back and forth

No screenshots bc it's backend code


apps/web/src/app/[locale]/settings/_components/settings.action.ts
```diff
@@ -25,18 +25,30 @@ export async function updateProfile(profileData: ProfileSchema) {
...
+  // Create an array of user links to create
+  const userLinksToCreate = profileData.userLinks.map((link) => ({
+    url: link.url,
+    user: {
+      connect: { id: session.user.id },
+    },
+  }));
+
+  // Insert the new user links in a batch
+  await prisma.$transaction(
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/settings/_components/settings.action.ts:50]: id put the delete and everything inside the transaction so if delete and create fails nothing will happen

PickleNik (approved):

apps/web/src/app/[locale]/settings/_components/settings.action.ts
Comment (maxdemaio) [reply apps/web/src/app/[locale]/settings/_components/settings.action.ts]: all set!

bautistaaa (approved):

PR #961, opened 2023-10-26
new(challenge): default-generic-arguments

dimitropoulos:


challenges/default-generic-arguments/prompt.md
```diff
@@ -0,0 +1,68 @@
...
+
+Technically speaking, no program absolutely requires default arguments.  It's a convenience feature for the humans writing programs.
+
+Sometimes you have a value in mind for an argument and you don't want to force users of your library to constantly provide the same value over and over again.  That's the situation where default arguments come in handy.
+
+## How To Create Default Type Arguments
+
+### First: A Refresher On Function Defaults
+
+Just like how you can create defaults arguments for functions.  Let's take the case of a run-of-the-mill logging function.  The logger always needs a message to log, but [log level](https://en.wikipedia.org/wiki/Syslog#Severity_level) might be something you want to be optional:
```
QuiiBz, line 11: Do we need the first sentence? It sounds a bit strange in this context
```suggestion
Let's take the case of a run-of-the-mill logging function.  The logger always needs a message to log, but [log level](https://en.wikipedia.org/wiki/Syslog#Severity_level) might be something you want to be optional:
```

QuiiBz (approved):
Nice!

challenges/default-generic-arguments/prompt.md
dimitropoulos: I think I just didn't delete enough when re-re-rewriting (which I do a lot of usually when making these, heh.)

PR #988, opened 2023-10-29
feat: added event for enroll / unenroll for tracks

### LINKED ISSUE #986: [Feature]: Add vercel event for enroll/unenroll button on /tracks
### Description

We current have an event for the submit button challenges. Lets add new ones for unenroll/enroll actions and log the track name as well. We currently have one track so it technically doesnt matter but as we scale tracks it will make sense.

### Explanation

_No response_

zaCKoZAck0:
## Description

Added event for enroll / unenroll for tracks tracking `slug` and action i.e. `enroll` and `unenroll`

## Related Issue

Closes #986


apps/web/src/app/[locale]/tracks/_components/track-details.tsx
```diff
@@ -50,6 +51,18 @@ export async function TrackDetail({ slug }: TrackDetailProps) {
     })
     .map((trackChallenge) => trackChallenge.id);
 
+  async function enroll(id: number) {
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/tracks/_components/track-details.tsx:54]: can we just move the track to the server actions that are defined already

bautistaaa (approved):

apps/web/src/app/[locale]/tracks/_components/track-details.tsx
Comment (zaCKoZAck0) [reply apps/web/src/app/[locale]/tracks/_components/track-details.tsx]: Ya I was initially doing it that way only but it was resulting in code changes in multiple places, below is the updated commit.

Issue #989, opened 2023-10-29
[Bug][Challenge] Index Signatures no proper test cases


niemal:
### What happened?

I passed all test cases with the following code (thing which shouldn't be the case):

```ts
type GroceryList = {
	[item: string]: number;
}
type InappropriateActionBySituation = unknown;

type CharactersById = unknown;
```
![](https://i.imgur.com/4zShC8K.png)

Link to the code with the challenge: https://typehero.dev/challenge/index-signatures?code=C4TwDgpgBA4gTgewMYTiAMgSwM7CgXigG8AoASAG1NgIBbALilzkwDsBzAXUdYFdaARqgDcJAL4lQkKAElWAQzBhEyzPJoBBJMEwJWAIRABlar3W7WBKL1YBrVggDurUZPDQAwgAt5cedtRsQxkAEysbeycXEiA

[...]

PR #991, opened 2023-10-29
fix: index-signatures challenge proper test cases

niemal:
## Description

Proper test cases.

## Related Issue

Closes [#989](https://github.com/typehero/typehero/issues/989)


dimitropoulos (changes requested):
unfortunately, in this case, our best bet is to duplicate the existing test code with `Expect<Equal<` blocks.  I really want to keep the "real world" examples, but below them all we can have the Expect/Equal situation with exactly the same stuff.

The reason doing what's here is not really optimal is you can still forcefully break the tests other ways, for example, you could also pass the test by setting all the types equal to `any`.

niemal:
> unfortunately, in this case, our best bet is to duplicate the existing test code with `Expect<Equal<` blocks. I really want to keep the "real world" examples, but below them all we can have the Expect/Equal situation with exactly the same stuff.
> 
> The reason doing what's here is not really optimal is you can still forcefully break the tests other ways, for example, you could also pass the test by setting all the types equal to `any`.

I forgot about the `any` case. I think the concept is we should intrigue the person to find out the solution without revealing it (`Expect<Equal<`). By using `NotEqual` we are just enforcing the user to think of the solution without having any means to actually view what the test directly wants.

Having an "under-the-hood" wrapper of any test for such type-check (unknown/any) is a way to deal with this but I believe the user reading the full type testing is more appropriate. In my next commit I move the `any`/`unknown` `NotEqual` to the bottom of the test case as they are not as important as the first ones.

dimitropoulos:
from our call:

```ts
export type Equal<X, Y> =
  (<T>() => T extends X ? 1 : 2) extends
  (<T>() => T extends Y ? 1 : 2) ? true : false;
export type Expect<T extends true> = Equal<T, true>;
export type Extends<T, U> = T extends U ? true : false;

type GroceryList = {
  [a: string]: number;
};

const groceryList: GroceryList = {
  carrots: 5,
  potatoes: 12,
  sweetPotatoes: 2,
  turnips: 1,
  parsnips: 1,
  beets: 10,
  radishes: 2,
  rutabagas: 1,
  onions: 3,
  garlic: 2,

  // @ts-expect-error intentionally invalid
  shouldError: "because it's a string",

  // @ts-expect-error intentionally invalid
  shouldAlsoError: true,
};

type testCaseGroceryListUnknown = Expect<Extends<
  string,
  keyof GroceryList
>>;

/** don't worry if if you don't understand this syntax, it's just checking the type of the GroceryList */
type GroceryListValues = GroceryList[keyof GroceryList];
type testCaseGroceryListUnknown2 = Expect<Equal<
  GroceryListValues,
  number
>>;
```

niemal:
I can't see how we can avoid using something like:
```ts
type testCaseCharactersById2 = Expect<NotEqual<CharactersByIdValues, any>>;
type testCaseCharactersById3 = Expect<NotEqual<CharactersByIdValues, unknown>>;
```

bautistaaa:
@dimitropoulos @niemal whats the status on this?

niemal:
> @dimitropoulos @niemal whats the status on this? id like to get this merged so others can be unblocked. not sure if it was fixed in the other pr adding solutions runner

Not sure either, waiting on @dimitropoulos.

Issue #993, opened 2023-10-29
[Feature]: Indicator a challenge has been solved on the challenge card


bautistaaa:
### Description

Currently, theres no way to tell you completed a challenge unless youre in the context of a track.

A couple things we should do

- [ ] change sorting of each category on explor from not done -> done
- [ ] profile page needs a completed section
- [ ] challenge UI needs some indicator that it has been solved
- [ ] explore cards needs some indicator that it has been solved

ja153903:
@bautistaaa Is this what you had in mind for the challenge UI?
![Screenshot 2023-10-29 at 9 24 18 PM](https://github.com/typehero/typehero/assets/20523874/8f1ffcec-c6c7-4ce0-9f1e-d08a9dfd5ea9)
![Screenshot 2023-10-29 at 9 24 10 PM](https://github.com/typehero/typehero/assets/20523874/1479f0b4-c163-47b3-a0f2-21a253e51462)

ja153903:
@bautistaaa For the completed section, do we want to do something like this on the profile page?

<img width="1484" alt="Screenshot 2023-10-31 at 11 27 43 AM" src="https://github.com/typehero/typehero/assets/20523874/293fc473-b7a5-4dd2-886f-a918cdc6cb1c">

bautistaaa:
yup!

jselmani:
Looks like this was completed in #1014.  Can we close this?

PR #997, opened 2023-10-29
feat: sort challenge by user submission

beaussan:
## Description
This change change the sorting of challenges by user submission, showing not done items first

## Related Issue
Solves the first part of #993

## How Has This Been Tested?
I would have love to add some unit test to this sort function, but I didn't see any way to do that in the repo. So I will rely on empiric testing for now.


apps/web/src/app/[locale]/explore/_components/explore.action.ts
```diff
@@ -30,6 +32,22 @@ export async function getChallengesByTagOrDifficulty(str: string, take?: number)
           }),
     },
     include: {
+      ...( session && session?.user?.id ? {
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/explore/_components/explore.action.ts:35]: i dont think you need the `session &&` since youre using optional chaining

apps/web/src/app/[locale]/explore/_components/explore-section.tsx
```diff
@@ -65,11 +65,22 @@ export async function ExploreSection({ title, tag, redirectRoute }: SectionProps
       <section className="relative flex w-full flex-col overflow-hidden rounded-[2.5rem]">
         <Carousel>
           {challenges
-            .sort((a, b) =>
-              difficultyToNumber[a.difficulty] !== difficultyToNumber[b.difficulty]
+            .sort((a, b) => {
+              const aHasSubmission = a?.submission?.length && a.submission.length > 0;
```
Comment (ja153903) [inline apps/web/src/app/[locale]/explore/_components/explore-section.tsx:69]: I think we can just do `a?.submission?.length > 0`.

Same thing for `bHasSubmission`

Comment (bautistaaa) [reply apps/web/src/app/[locale]/explore/_components/explore-section.tsx:69]: you'd have to `(a?.submission?.length ?? 0) > 0` which isnt much better tbh

Comment (ja153903) [reply apps/web/src/app/[locale]/explore/_components/explore-section.tsx:69]: that's fair I guess. I was thinking more along the lines of `undefined > 0` will end up being false anyway.

Comment (bautistaaa) [reply apps/web/src/app/[locale]/explore/_components/explore-section.tsx:69]: TS wont let you do that

apps/web/src/app/[locale]/explore/_components/explore.action.ts
Comment (beaussan) [reply apps/web/src/app/[locale]/explore/_components/explore.action.ts]: Changed ! I thought the id was optional on a user

apps/web/src/app/[locale]/explore/_components/explore-section.tsx
Comment (beaussan) [reply apps/web/src/app/[locale]/explore/_components/explore-section.tsx:69]: Indeed, and imo, I rather have the double check than using the `??` operator here

bautistaaa:
@beaussan this is still in draft. now sure if you wanted to get this merged

beaussan:
@bautistaaa it should be good to go now :)

beaussan:
Hold on, somone added the summision in the action code, changing to reflect this

beaussan:
Done ! Should be good to go now

bautistaaa (approved):

bautistaaa:
@beaussan i dont think its working in prod 👀 

deployed it a bit ago but dont see the sorting taking effect.

beaussan:
@bautistaaa it seems like the sorting order is reversed :eyes: 

I'll make a small unit test to debug it further, I won't add it to the project since there is no unit test setup currently

PR #1000, opened 2023-10-30
Adding Check Icon to Challenge description as Solved indicator

ja153903:
## Description

See PR title

## Related Issue

Closes checkbox 3 on https://github.com/typehero/typehero/issues/993#issue-1967196007

Screenshot 2023-10-29 at 9 34 51 PM
Screenshot 2023-10-29 at 9 34 40 PM


bautistaaa:
cc @PickleNik thoughts?

bautistaaa:
Can you add a tooltip as to what this represents. Could just be "Solved"

apps/web/src/app/[locale]/challenge/[slug]/getChallengeRouteData.ts
```diff
@@ -35,6 +35,11 @@ export const getChallengeRouteData = cache(async (slug: string, session: Session
           userId: session?.user.id || '',
         },
       },
+      submission: {
+        where: {
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/challenge/[slug]/getChallengeRouteData.ts:39]: we can probably just take 1 thats successful via prisma and void needing to use some below

ex: where isSuccessful is true (i forgot the exact column name)

Comment (ja153903) [reply apps/web/src/app/[locale]/challenge/[slug]/getChallengeRouteData.ts:39]: Sure that makes sense. If I make this change, do you want me to move the check for hasSuccessfulSubmission over to the component using it instead of here?

Comment (ja153903) [reply apps/web/src/app/[locale]/challenge/[slug]/getChallengeRouteData.ts:39]: @bautistaaa I removed the part where I augment the `challenge` property. So we can just check length in the `Description` component.

PickleNik:
> cc @PickleNik thoughts?

me likey

apps/web/src/app/[locale]/challenge/[slug]/getChallengeRouteData.ts
Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/[slug]/getChallengeRouteData.ts:39]: we need to keep the explicitness of what having a successful solution is

submissions length > 0 meaning will be lost over time.

for now i guess you can do 

```ts
const hasSolved = challenge.submission.length > 0
```

or something in the component below or return it explicitly here

bautistaaa (approved):

PR #1009, opened 2023-10-30
adds solution runner

### LINKED ISSUE #1019: [Bug]: Test cases in Mapped Object Types don't allow you to pass the tests
### What happened?

https://typehero.dev/challenge/mapped-object-types :(

_No response_

### Relevant log output

_No response_

dimitropoulos:
this is a draft PR to try and get a solution runner going so we are never in the situation where tests are broken or solutions are not possible.

closes #1019


package.json
```diff
@@ -32,10 +33,14 @@
...
     "markdownlint-cli2": "^0.10.0",
+    "picocolors": "^1.0.0",
     "prettier": "^3.0.2",
     "prettier-plugin-tailwindcss": "^0.4.1",
     "turbo": "^1.10.16",
-    "typescript": "^5.1.6"
+    "tsx": "^3.14.0",
+    "type-testing": "0.2.0",
+    "typescript": "^5.1.6",
+    "vitest": "^0.34.6"
```
Hacksore, line 43: i tried playing with vitest for testing types but it's a lot of work haha

Hacksore:
We either need to copypasta the testing types into the code or find a way to make them work within the typescript code.

Current output:
```

> typehero@0.1.0 check:solutions /Users/hacksore/code/opensource/typehero
> tsx ./challenges/validate.ts

✗ default-generic-arguments
Cannot find module 'type-testing' or its corresponding type declarations.

✗ generic-function-arguments
Cannot find module 'type-testing' or its corresponding type declarations.
Block-scoped variable 'identity' used before its declaration.
Block-scoped variable 'identity' used before its declaration.
Block-scoped variable 'identity' used before its declaration.
Block-scoped variable 'mapArray' used before its declaration.
Block-scoped variable 'mapArray' used before its declaration.
Block-scoped variable 'mapArray' used before its declaration.
Block-scoped variable 'mapArray' used before its declaration.

✗ generic-type-arguments
Cannot find module 'type-testing' or its corresponding type declarations.
Cannot find name 'GroceryItem'.

✗ generic-type-constraints
Cannot find module 'type-testing' or its corresponding type declarations.

✓ index-signatures


✗ indexed-types
Cannot find module 'type-testing' or its corresponding type declarations.
Tuple type 'Cars' of length '5' has no element at index '5'.

✓ keyof


✗ literal-types
Cannot find module 'type-testing' or its corresponding type declarations.

✗ mapped-object-types
Cannot find module 'type-testing' or its corresponding type declarations.

✗ pick
Cannot find name 'MyPick'. Did you mean 'Pick'?
Cannot find name 'MyPick'. Did you mean 'Pick'?
Duplicate identifier 'Pick'.

✗ primitive-data-types
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'artistName'.
Cannot redeclare block-scoped variable 'name'.
Type 'void' is not assignable to type 'string'.

✗ type-aliases
Cannot find name 'Expect'.
Cannot find name 'Equal'.
Cannot find name 'Expect'.
Cannot find name 'Equal'.
Cannot find name 'Expect'.
Cannot find name 'Equal'.
Cannot find name 'Expect'.
Cannot find name 'Equal'.
Cannot find name 'Expect'.
Cannot find name 'Equal'.
Cannot find name 'Expect'.
Cannot find name 'Equal'.
Cannot find name 'Expect'.
Cannot find name 'Equal'.

✗ type-unions
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'Expect'.
Cannot find name 'Expect'.

✗ typeof
Cannot find module 'type-testing' or its corresponding type declarations.
```

dimitropoulos:
Screenshot_20231031_140216

got 'em all passing!

Hacksore:
@dimitropoulos we need to think about sharing the same `tsconfig.json` between the client and test runner 😅.

challenges/validate.ts
```diff
@@ -91,4 +148,142 @@ const validateMetadataFiles = () => {
...
+
+      const testsSource = readFileSync(join(challengesDir, id, 'tests.ts'), 'utf8');
+
+      readdirSync(join(challengesDir, id, 'solutions')).forEach((file) => {
+        const solutionPath = join(challengesDir, id, 'solutions', file);
+        const solutionSource = readFileSync(solutionPath, 'utf8');
+
+        // our challenge test file
+        const sourceFile = createSourceFile(
+          `${repoRoot}/in-memory/${file}`,
```
Hacksore, line 181: in the future we should split them up into two files `user.ts` and `test.ts` would help narrowing down the error lines in output easier.

I was playing with that here
https://github.com/Hacksore/typecheck-service/blob/feat/worker-cf/src/server.ts#L52-L53

dimitropoulos: I was thinking that too, but the problem there is we'd need to insert our own imports which will throw off line counts.  beyond that, for imports we'd need to export things which is probably a bigger issue because then we'd need to go in and find all the symbols the user may have created.  or am I missing something?

Hacksore: agreed it's not required now let me resolve this.

dimitropoulos:
I think this is at least close to something mergeable.  At a least it fixes some fairly significant problems with various challenges.  We really should find a way to integrate this into CI or something, but I'd tend to say we should merge it prematurely to prevent all the "this challenge is broken" situations from occurring.

Hacksore (approved):

PR #1012, opened 2023-10-31
fix: Next 14

ctjlewis:
Don't tell them I said it, but server components are still pretty fragile. Avoid this pattern for now:

```ts
// some non-component
import { utility } from "./server-component"
```


ctjlewis:
Btw, this technically breaks the rules as all the exports from `track.action.ts` should be async functions,. I just couldn't bring myself to create a new file for two one-liners, and it wasn't throwing; we can't remove the one-liners or it's a DRY issue; and I can't bring myself to pointlessly await synchronous string templates either.

Should do something there before shipping but wanted to get this in as a draft to start.

lubieowoce:
> Btw, this technically breaks the rules

i guess one alternative would be to use an inline "use server" (within each action) instead of a top-level one for the whole file? then, exporting a non-async function from the same file should be fine, it won't be considered an action or anything

PR #1034, opened 2023-11-02
fix: sorting order was reversed on completed challanges

beaussan:
## Description
Sorting order was reversed, see https://github.com/typehero/typehero/pull/997#issuecomment-1789841797 

## How Has This Been Tested?
By [unit tests, in this stackblitz](https://stackblitz.com/edit/vitest-dev-vitest-ejxeya?file=src%2Fsort.test.ts) since there is no unit test setup in the project.


beaussan:
@bautistaaa PR up with the fix ! There was no unit test in place, so I didn't wanted to force it on the codebase, but I've provided unit tests in a sandbox to make sure the functionality was working properly now ;)

bautistaaa (approved):

Issue #1031, opened 2023-11-01
[Bug]: Kilograms and Count are missing Type Alias challenge


is-it-ayush:
### What happened?

The following types `Kilograms` and `Count` are missing on [type-aliases challenge](https://typehero.dev/challenge/type-aliases). This should be mentioned in the challenge description under "Solve the challenge" section.

### What browsers are you seeing the problem on?

All

dimitropoulos:
Did you see the section that reads:

```
  // the tests show that you need a `mass` property here
  // but first you might need to create an alias for `Kilograms`
  // because that's the value of `mass`
```

The idea is to give you some practice actually creating an alias from scratch, which for this one challenge means we can't pre-write it for you if the goal is for it to really give you some practice making type aliases.  So, in this case it's intended behavior for you to have to write `type Kilograms = `.  Does that make sense?

is-it-ayush:
I did read that @dimitropoulos! This issue primarily was for `Count` although I included `Kilograms` since it was also only mentioned in the comments. `Count` isn't mentioned anywhere. I think it'd be better to mention them in the challenge description so one doesn't have to search the UI for what to write from a user's perspective who is expected to be new to typescript. This is suggestion so if it isn't something that is planned. We could close this issue!

dimitropoulos:
@is-it-ayush although it's the intended behavior, it's still really helpful feedback!  thanks!  see the attached PR and lemme know what you think :)

bautistaaa:
going to close since we fixed the languag

PR #1039, opened 2023-11-02
[Feature] challenge(`type-alias`) improves the prompt

dimitropoulos:
Improves the prompt of the type-alias challenge to signal to users that they might be asked to create a type alias.

## Description

It may be confusing to see some type aliases pre-created and have some missing.  Indeed, in this specific challenge this is very much intended behavior because we want users to get some on-the-ground experience with writing type-aliases from scratch.

## Related Issue

Closes # https://github.com/typehero/typehero/issues/1031

thanks again @is-it-ayush for bringing this up

### DIFF
--- a/challenges/type-aliases/prompt.md
+++ b/challenges/type-aliases/prompt.md
@@ -85,6 +85,8 @@ If the answer is yes, then you should probably not make an alias for `Rows`. Thi
 
 ## Solving This Challenge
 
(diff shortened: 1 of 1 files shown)


bautistaaa (approved):
i like it

is-it-ayush:
thank you @dimitropoulos!! that certainly looks good. 💙✨

PR #1046, opened 2023-11-03
Form Message Color Contrast

### LINKED ISSUE #1029: [Bug]: Not enough color contrast for the 'Too short!' error message in comments section of a type challenge
### What happened?

More of an accessibility bug.

The error message that comes up when the comment box is empty is too dark for the background which would make it difficult for users with color blindness issues to see the message.

Here an example of the bug:

Screenshot from 2023-11-01 17-43-39

Chrome

### Relevant log output

```zsh
N/A
```

markkhoo:
## Description

This gives the **form message** in `form.tsx` more color contrast. Particularly, the form message when choosing a title for a solution meets a `4.5:1` color contrast ratio in dark mode. It does not meet this ratio for the message when leaving a comment in dark mode. Light mode has it's own issues... Anyone have any thoughts?

## Related Issue

Closes #1029


chadstewart:
Hey @bautistaaa, @PickleNik. Have any input on this?

bautistaaa (approved):

PR #1065, opened 2023-11-07
fix(web): i18n locales error due to toolbar

### LINKED ISSUE #1064: [Bug]: invalid routes causing runtime errors
### What happened?

example route: https://typehero.dev/dcgt/dcff

image
image

_No response_

### Relevant log output

_No response_

QuiiBz:
## Description

`@vercel/toolbar` makes a call to `/.well-known/vercel-user-meta` which seems to bypass `generateStaticParams` so we get a `.well-known` dynamic locale segment, which doesn't exist.

## Related Issue

Closes #1064

## How Has This Been Tested?

- `cd apps/web && vc link`
- `pnpm dev` then navigate to the homepage, there shouldn't be any error logged in the terminal

UPDATE:
Only bump next-international to include https://github.com/QuiiBz/next-international/pull/252

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -52,7 +52,7 @@
     "million": "2.6.0-beta.12",
     "monaco-vim": "^0.4.0",
     "next": "14.0.1",
-    "next-international": "1.1.2",
+    "next-international": "1.1.4",
     "next-themes": "^0.2.1",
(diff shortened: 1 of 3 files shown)


bautistaaa (approved):
thanks for the quick fix <3

PR #1080, opened 2023-11-09
chore: move user route

bautistaaa:
Remove the ambiguous route in hope to fix the odd 404 issues


zachlankton:
This PR should not be needed if PR #1084 is accepted and merged

PR #1084, opened 2023-11-09
Fix 404 frfr

zachlankton:
## Description
Fix 404 for real this time

Related #1080 

## How Has This Been Tested?
Manually in pnpm dev 
Also Manually with build and start

image


zachlankton:
@QuiiBz would you mind taking a look at how we are doing this with the I18nProviderClient component in the [locale]/layout.tsx ?  Curious if you think this is a good idea or not... basically everything here is geared toward getting 404 and not-found.tsx rendering properly.

apps/web/src/app/layout.tsx
```diff
@@ -0,0 +1,78 @@
...
+        />
+        <link href="/favicon.svg" rel="icon" type="image/svg+xml" />
+        <link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png" />
+        <link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png" />
+        <link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png" />
+        <link rel="manifest" href="/site.webmanifest" />
+      </head>
+      <body className={`${inter.className} flex flex-col`}>
+        <Providers>
+          <Navigation />
```
QuiiBz, line 70: Doing this means we won't be able to use i18n inside `<Navigation />` and its childrens :(

QuiiBz (changes requested):

apps/web/src/app/layout.tsx
zachlankton: Excellent, thank you! refactored to put Nav back in [locale]

apps/web/src/app/[locale]/layout.tsx
```diff
@@ -62,26 +10,11 @@ export default function RootLayout({
...
-          <Toaster />
-        </Providers>
-        <Analytics />
-      </body>
-    </html>
+    <>
+      <Navigation />
+      <main className="flex-1">
+        <I18nProviderClient locale={locale}>{children}</I18nProviderClient>
+      </main>
```
Comment (QuiiBz) [inline apps/web/src/app/[locale]/layout.tsx:17]: Shouldn't `Navigation` be wrapped inside `I18nProviderClient`?

QuiiBz (changes requested):

apps/web/src/app/[locale]/layout.tsx
Comment (zachlankton) [reply apps/web/src/app/[locale]/layout.tsx:17]: Really not sure. In your example all nav is in the layout.tsx and the provider is used to wrap some client components that dont have any nav? 

Or am i missing it?

I can try it tho and see what happens.

Comment (zachlankton) [reply apps/web/src/app/[locale]/layout.tsx:17]: It seems to work... is this the recommended approach? Why is the app route example different?

Comment (bautistaaa) [reply apps/web/src/app/[locale]/layout.tsx]: thats weird that it works not being nested 😅

Comment (zachlankton) [reply apps/web/src/app/[locale]/layout.tsx]: confirmed Navigation needs to be inside the provider.... the reason it worked before was we weren't doing any i18n inside the nav component.  Once I added i18n the component would break if it was outside the context provider

Comment (QuiiBz) [reply apps/web/src/app/[locale]/layout.tsx]: Yeah exactly this ^

QuiiBz (approved):

apps/web/src/app/not-found.tsx
```diff
@@ -0,0 +1,75 @@
+import Image from 'next/image';
+import { Button } from '@repo/ui/components/button';
+import { getRandomChallenge } from '~/utils/server/get-random-challenge';
+import { getScopedI18n } from '~/locales/server';
+import { setStaticParamsLocale } from 'next-international/server';
+
+export default async function NotFound() {
+  // fix broken e2e tests... not sure why this is needed
+  // but i18n still works with other locales even with this set
+  setStaticParamsLocale('en');
```
zachlankton, line 10: actually just discovered this is not true.... stand by, looking at how to fix.

zachlankton: @QuiiBz I'm getting the following error only during pnpm build (pnpm dev works great)... when I add the code above I can supress the error and build succeeds, but breaks i18n on the 404 not-found.tsx page.  not-found.tsx is in the root app folder it does not have access to the `locale` parameter that we do in pages nested under `[locale]`

```bash
Error occurred prerendering page "/_not-found". Read more: https://nextjs.org/docs/messages/prerender-error
Error: Could not find locale while pre-rendering page, make sure you called `setStaticParamsLocale` at the top of your pages
    at /home/zach/typehero/apps/web/.next/server/chunks/4964.js:1:150920
    at /home/zach/typehero/node_modules/next/dist/compiled/next-server/app-page.runtime.prod.js:108:5119
    at /home/zach/typehero/apps/web/.next/server/chunks/4964.js:1:152590
    at Generator.next (<anonymous>)
    at /home/zach/typehero/apps/web/.next/server/chunks/4964.js:1:150131
    at new Promise (<anonymous>)
    at __async (/home/zach/typehero/apps/web/.next/server/chunks/4964.js:1:149909)
    at /home/zach/typehero/apps/web/.next/server/chunks/4964.js:1:152554
    at NotFound (/home/zach/typehero/apps/web/.next/server/chunks/7386.js:1:10186)

```

QuiiBz: We might have to opt-out of static rendering for this page. `setStaticParamsLocale` is used when building to retrieve the correct locale, so it should be used in this page (even if we hardcode `"en"`)

bautistaaa: > We might have to opt-out of static rendering for this page. `setStaticParamsLocale` is used when building to retrieve the correct locale, so it should be used in this page (even if we hardcode `"en"`)

so we need to add this to the top?

```ts
export const dynamic = 'force-dynamic';
```

QuiiBz: Just tried and `pnpm dev` works fine for 404 pages, and `pnpm build` also succeed

zachlankton: Excellent, I tested also, works great, just pushed change!

bautistaaa (approved):

bautistaaa:
🐐

Issue #1077, opened 2023-11-08
[Bug]: Test panel should be open by default on challenges


bautistaaa:
### What happened?

A bug happened!

rbrtj:
Sorry for not asking to assign me for the issue but i was just exploring the repo and decided to contribute.

PR #1089, opened 2023-11-10
fix: tests panel open by default on challenges

rbrtj:
## Description

I've only changed the initial state, so the test panel is open by default when it's not a playground (I didn't really find any other views using the editor).
## Related Issue
[#1077](https://github.com/typehero/typehero/issues/1077)
Closes #

Challenge view:
image
Playground view:
image


packages/monaco/src/index.tsx
```diff
@@ -40,7 +40,7 @@ export function CodePanel(props: CodePanelProps) {
   const isPlayground = pathname.includes('playground');
   const { toast } = useToast();
   const [tsErrors, setTsErrors] = useState<TsErrors>();
-  const [isTestPanelExpanded, setIsTestPanelExpanded] = useState(false);
+  const [isTestPanelExpanded, setIsTestPanelExpanded] = useState(!isPlayground);
```
bautistaaa, line 43: can just use `true`

rbrtj: I thought that it shouldn't be opened by default in the playground view, but if that is not the case, then sure, I'll change it to true.

bautistaaa (approved):

Issue #1104, opened 2023-11-12
[Bug]: Track UI on challenge view pushes comments UI offscreen


bautistaaa:
### What happened?

Comment section is not account for the extra space the tracks ui occupies at the top of the panel so its offset

https://typehero.dev/challenge/generic-function-arguments

unable to scroll down farther to see then bottom
![image](https://github.com/typehero/typehero/assets/3660667/a1de897e-eb98-460f-855d-d0f910e1b43c)

aadamw:
On it

me-imfhd:
working on it... assign me this issue

bautistaaa:
updated description and DM'd you a token @me-imfhd

PR #1113, opened 2023-11-13
fix: comments ui

me-imfhd:
will ensure that when comment section is open, it will start the scroll from top
## Related Issue

Closes https://github.com/typehero/typehero/issues/1104


me-imfhd:
latest commit fixes comments bottom issue
[Screencast from 14-11-23 12:33:59 PM IST.webm](https://github.com/typehero/typehero/assets/114667178/8b0d9f7f-ddb5-43c2-96ee-ebd83e3cf2da)

me-imfhd:
please ignore the alert, it shows whether isVisible state is true or false when track title header is there or not there,
i have removed it in one commit....

also i have used context api for state management just around the challenge slug route

[Screencast from 15-11-23 02:42:12 AM IST.webm](https://github.com/typehero/typehero/assets/114667178/98735c18-9cb7-4f5a-af10-1fd6b0bce5cc)

aadamw:
I think the solution is overcomplicated. I think there is no need to use context we can check if `track` is not undefined inside `challenge/[slug]/page.tsx` and pass an additional prop isTrackView to `<Comments />` component and adjust styling based on that.
You can extract `track` data from [getChallengeRouteData()].(https://github.com/typehero/typehero/blob/main/apps/web/src/app/%5Blocale%5D/challenge/%5Bslug%5D/page.tsx#L38)

me-imfhd:
Hey @aadamw thanks for reviewing this,
The reason I have used state is because whether it is visible or not does not just depend on track being defined or undefined it also depends on other factors 

Check this screenshot and read the isTrackVisible field,

This tells whether that track header will be visible or not not several other factors
Thus i have made the state directly depend on this,
And used the context to avoid prop drilling

The solution you proposed will only work when the track is just defined or not, but would be buggy and most probably wouldn't work in other cases

 But please let me know if i am missing something, i have just started to familiarize myself with the code....

apps/web/src/app/[locale]/challenge/_components/comments/index.tsx
```diff
@@ -79,9 +83,11 @@ export function Comments({ preselectedCommentMetadata, rootId, type, expanded =
           className={clsx(
             'custom-scrollable-element flex flex-col overscroll-contain duration-300',
             {
-              'h-64 pb-4 md:h-[calc(100vh_-_164px)]': showComments,
               'h-0 overflow-y-hidden': !showComments,
-              'overflow-y-auto': showComments && (data?.comments.length ?? 0) > 0,
+              'h-64 pb-4 lg:h-[calc(100vh_-_164px)]': showComments && !isTrackTitleVisible, // no title
+              'h-[calc(256px_-_46px)] pb-4 lg:h-[calc(100vh_-_210px)]':
+                showComments && isTrackTitleVisible, // with title
+              'overflow-y-auto ': showComments && (data?.comments.length ?? 0) > 0,
```
Comment (ayrock-dev) [inline apps/web/src/app/[locale]/challenge/_components/comments/index.tsx:90]: thought: This looks a little dense, @me-imfhd.It might be brittle to keep in-sync. For example, that `210px` needs to be maintained as the exact height of the title and header.

I'm wondering if there is a simpler way. If this element is not meant to be `100vh`, I wonder if we should not be using `100vh` in the calculation. Instead, maybe a parent/ancestor can be `100vh` and the comments drawer here can instead use `height: 100%` to fill the size of it's parent/ancestor.

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
```
Comment (ayrock-dev) [inline apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:1]: praise: this file looks great!

(non-blocking) thought: I wonder if there is a better name for this file, since `context` is an implementation detail for `useTrackVisibility`. Instead i might name this file with `hooks` in the name, or, `use-track-visibility.tsx` as it is tied to a single hook.

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
```diff
@@ -0,0 +1,29 @@
...
+  isTrackTitleVisible: boolean | null;
+  setIsTrackTitleVisible: React.Dispatch<React.SetStateAction<boolean | null>>;
+}
+
+export const TrackVisibilityContext = createContext<TrackVisibilityContextType | undefined>(
+  undefined,
+);
+
+export const TrackVisibiltyProvider = ({ children }: PropsWithChildren) => {
+  const [isTrackTitleVisible, setIsTrackTitleVisible] = useState<boolean | null>(null);
```
Comment (aadamw) [inline apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:14]: Question: Why can't we have just boolean over there with initial value of `false`?

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
```diff
@@ -0,0 +1,29 @@
+'use client';
+import { createContext, useState, useContext, type PropsWithChildren } from 'react';
+
+interface TrackVisibilityContextType {
+  isTrackTitleVisible: boolean | null;
+  setIsTrackTitleVisible: React.Dispatch<React.SetStateAction<boolean | null>>;
+}
+
+export const TrackVisibilityContext = createContext<TrackVisibilityContextType | undefined>(
```
Comment (aadamw) [inline apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:9]: I think we don't need `undefined` there as we will always have those values when our components are wrapped with the provider

```suggestion
export const TrackVisibilityContext = createContext({} as TrackVisibilityContextType)
```

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <TrackVisibilityContext.Provider value={{ isTrackTitleVisible, setIsTrackTitleVisible }}>
+      {children}
+    </TrackVisibilityContext.Provider>
+  );
+};
+
+export const useTrackVisiblity = () => {
+  const context = useContext(TrackVisibilityContext);
+  if (context === undefined) {
+    throw new Error('use the context within context provider');
```
Comment (aadamw) [inline apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:26]: non-blocking:
```suggestion
    throw new Error('use useTrackVisiblity hook within TrackVisibiltyProvider');
```

apps/web/src/app/[locale]/challenge/_components/comments/index.tsx
```diff
@@ -23,8 +24,10 @@ interface Props {
 
 // million-ignore
 export function Comments({ preselectedCommentMetadata, rootId, type, expanded = false }: Props) {
+  const { isTrackTitleVisible } = useTrackVisiblity();
   const [showComments, setShowComments] = useState(expanded);
   const commentContainerRef = useRef<HTMLDivElement>(null);
+  // const tabsRef = useRef<>(null);
```
Comment (aadamw) [inline apps/web/src/app/[locale]/challenge/_components/comments/index.tsx:30]: Leftover comment

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
```diff
@@ -0,0 +1,29 @@
...
+  const [isTrackTitleVisible, setIsTrackTitleVisible] = useState<boolean | null>(null);
+
+  return (
+    <TrackVisibilityContext.Provider value={{ isTrackTitleVisible, setIsTrackTitleVisible }}>
+      {children}
+    </TrackVisibilityContext.Provider>
+  );
+};
+
+export const useTrackVisiblity = () => {
```
Comment (aadamw) [inline apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:23]: suggestion:
I think this name would be more informative. WDYT? Same for context, provider and state names
```suggestion
export const useTrackNavigationVisiblity = () => {
```

aadamw (commented):
Thanks for working on it @me-imfhd. I'm still learning the codebase and wasn't aware of some things. I left few small suggestions from my side that could improve readability

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
Comment (bautistaaa) [reply apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:14]: lets default to false

bautistaaa:
@me-imfhd can you address the comments please so we can merge

apps/web/src/app/[locale]/challenge/_components/comments/index.tsx
Comment (me-imfhd) [reply apps/web/src/app/[locale]/challenge/_components/comments/index.tsx:90]: hey @ayrock-dev thanks for reviewing, i tried the simpler way but it didn't work out as i expected, reason was the way components were written or just maybe i am not that skilled yet...
and the other thing that
yeah that 210px is not dynamic, but for the time being since everything was static so the code was quite working for everything so i didn't wanted to increase the code logic thus i stick with this, but maybe this can be written better...

By the way, my reason to write h-[calc(256px_-_46px)] this instead of h-[210px] was that h-64 is equal to 256px and thus we know reducing 46px from h-64 will make it work, maybe it was not good idea but that was my resolve.
But i can just write h-[210px] if its asked.

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
Comment (me-imfhd) [reply apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:1]: yeah making this change

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
Comment (me-imfhd) [reply apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:14]: okay

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
Comment (me-imfhd) [reply apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:9]: oh thanks for the code i didn't know i could get rid of setting initial value like this

apps/web/src/app/[locale]/challenge/_components/comments/index.tsx
Comment (me-imfhd) [reply apps/web/src/app/[locale]/challenge/_components/comments/index.tsx:30]: removed

apps/web/src/app/[locale]/challenge/[slug]/_context.tsx
Comment (me-imfhd) [reply apps/web/src/app/[locale]/challenge/[slug]/_context.tsx:23]: W naming

me-imfhd:
@bautistaaa are we good?

bautistaaa (approved):
trusting this works

me-imfhd:
😢 i hope the best !

Issue #1117, opened 2023-11-14
[Feature][Advent of TypeScript 2023]: Create Track


ayrock-dev:
#### References
 - https://github.com/typehero/typehero/discussions/1078

#### Requirements
 - Add Advent of TypeScript track. Be sure to use the year (2023) in the title for the track.
 - Slug should be `aot-2023`
 - Update seed scripts to create this track

![image](https://github.com/typehero/typehero/assets/7817695/50e42c82-efaf-4040-95f2-33c84264cd5f)

PR #1126, opened 2023-11-15
feat: aot: seed local track

ayrock-dev:
## Description

seed a local Advent of TypeScript track so folks can use this

TODO
 - [x] consider if we need a hard-coded slug or if its fine that it's automatically generated (`slugify()`) by the track title

edit: this ^ is looking like a non-concern!

## Related Issue

Part of #1117


bautistaaa:
slug name wont matter since we'll only be using it in a query from `aot-2023` route

bautistaaa (approved):

Issue #1106, opened 2023-11-12
[Bug]: Track UI on challenge view seems to be off by 1


bautistaaa:
### What happened?

navigate to second track and you cannot hit the previous button

me-imfhd:
can i know on which route is this issue arising... how to see it ?

PR #1129, opened 2023-11-15
fix: navigate to second track and you cannot hit the previous button

me-imfhd:
made two small changes:
1) In previous index should return the track challenge also if its zero and not just greater than zero
2) named index to prevIndex and nextIndex for better readability  
## Description

`return index > 0 ? trackDetails!.trackChallenges[index] : null;`
this will return null also if the index is 0 but since indexing starts from 0 
so the code should look like
`return index >= 0 ? trackDetails!.trackChallenges[index] : null;`

## Related Issue

https://github.com/typehero/typehero/issues/1106
Closes #


zaCKoZAck0:
Looks great.

zaCKoZAck0 (approved):

PR #1137, opened 2023-11-17
feat: aot track icon

ayrock-dev:
Add tree icon for various track iconography. Can replace icon with whatever we want or something better later.

Setting this up as example of reading track slug to show some extra fun visuals.

|Track Card on `/tracks`|
|---|
|image|

|Track Detail|
|---|
|image|

|In-Challenge Track Icon|
|---|
|image|


bautistaaa:
i don’t think they’ll be able to see the track ui in challenge view cause we’re just revealing them per day. or are you envisioning if they are on day 5 they can cycle through those 5?

Track details will need go to /aot-2023. we can change link based on track name methinks

ayrock-dev:
That makes sense to me. I had just done a cmd+shift+F for where we render the <Swords /> icon. I think you are right, most of these will not be visible since it's a hidden track.

Maybe I can do a follow-up to hide some stuff, I think the in-challenge track overview will need to take you to /aot-2023 instead of the track overlay as well.

ayrock-dev:
closing, we will manage track visibility

PR #1175, opened 2023-11-23
Fix ordered list display in markdown

### LINKED ISSUE #1172: [Bug]: Type Union markdown not properly markeddown
### What happened?

It displays like this: 
Screenshot 2023-11-23 103640

At https://typehero.dev/challenge/type-unions 

But the .md looks correct to me: https://github.com/typehero/typehero/blob/main/challenges/type-unions/prompt.md#properties-of-unions

Chrome

### Relevant log output

_No response_

heathhenley:
Try to fix order lists, I think tailwinds default removes list styles and typography `prose` is supposed to put it back, but the description is using `prose-invert` with no `prose`. So I just put the styles from `prose` that are relevant to lists in `prose-invert` - not sure if there's actually a better way to do it?

## Description
- Copy the .prose styles from typography to .prose-invert so that the order lists display with the numbers in challenge prompts.
- Edit type-unions to remove sub indent in list (the `a. ... ` elements weren't not parsing as `<li>`'s, may not actually be supported by react-markdown 

## Related Issue
Closes #1172 

## How Has This Been Tested?
I ran e2e tests locally, the same results as main

Before: 
pick-list-old

After:
pick-list


bautistaaa:
can we apply the correct classes here?

https://github.com/typehero/typehero/blob/main/packages/ui/src/components/markdown.tsx#L49-L64

heathhenley:
That makes a lot more sense, thanks for pointing that out. Want me to squash the last two together?

bautistaaa (approved):

PR #1192, opened 2023-11-27
fix(e2e-test): add flaky test (`comment.test.ts`) by adding timeout

### LINKED ISSUE #1180: [Bug]: Fix e2e tests
### What happened?

they broke when we refactored the comments section

_No response_

### Relevant log output

_No response_

DerTimonius:
## Description

This PR fixes the previously failing e2e test for comments by adding a timeout. I noticed that the deletion of the comment takes some time to get reflected in the frontend.

## Related Issue

Closes #1180

## How Has This Been Tested?

Test runs without failure locally (hopefully also in the CI)

### DIFF
--- a/apps/web/tests/e2e/authenticated/comment.test.ts
+++ b/apps/web/tests/e2e/authenticated/comment.test.ts
@@ -97,7 +97,7 @@ test.describe('create, edit, and delete comments', () => {
 
     await expect(
       page.getByLabel('Description').getByText(`${editedParentComment} ${parentCodeText}`),
-    ).not.toBeVisible();
+    ).not.toBeVisible({ timeout: 10000 });
   });
 });
 


apps/web/tests/e2e/authenticated/comment.test.ts
```diff
@@ -97,7 +97,7 @@ test.describe('create, edit, and delete comments', () => {
 
     await expect(
       page.getByLabel('Description').getByText(`${editedParentComment} ${parentCodeText}`),
-    ).not.toBeVisible();
+    ).not.toBeVisible({ timeout: 10000 });
```
bautistaaa, line 100: this hurts my soul but welcome to e2e i guess.

is there a global timeout config we can put somewhere so it just does this on any of these?

DerTimonius: It could be done according to the [docs](https://playwright.dev/docs/test-timeouts#expect-timeout), would this be your preferred method?

bautistaaa: yeah lets do that instead cause i think we'll end up with a lot of ad hoc timeouts otherwise. thoughts?

DerTimonius: I personally like the approach of adding the timeouts as needed better, because I feel that sometimes the global timeouts are not working as expected. Although I only set the _test_ timeout before, not the _expect_ timeout, so maybe this config would be more reliable?

I will try that later/tomorrow

bautistaaa: > I personally like the approach of adding the timeouts as needed better, because I feel that sometimes the global timeouts are not working as expected. Although I only set the _test_ timeout before, not the _expect_ timeout, so maybe this config would be more reliable?
> 
> I will try that later/tomorrow

ok lets merge this. we can address if it becomes more of an issue down the road!

bautistaaa (approved):

PR #1197, opened 2023-11-29
The text color of the solution button should remain white.

roiiiu:
## Description

image

The overview page `Solution` button text is not very readable in light mode.

image

## Related Issue

Closes #


bautistaaa (approved):
thanks!

PR #1212, opened 2023-11-30
fix: patch next auth

bautistaaa:


patches/@auth__core@0.0.0-manual.e9863699.patch
Hacksore, line 1: idk why this diff is massive it should be one line patch but approving anyway!

Hacksore (approved):

patches/@auth__core@0.0.0-manual.e9863699.patch
bautistaaa: i think prettier formatting

PR #1222, opened 2023-12-01
fix: Fix typos in literal-types challenge

romain-trotard:
Fix some typo in the `literal-types` challenge.

## Description

When doing the challenge I have seen some improvements.

- The first we can see in the description of the challenge:

image

- The other fix is that there is no `almostPi` variable in the user panel.

## Related Issue

I don't have created an issue for this fixes. But if I should do, tell me ;)

Thanks for the site <3

### DIFF
--- a/challenges/literal-types/user.ts
+++ b/challenges/literal-types/user.ts
@@ -7,6 +7,5 @@ type LiteralFunction = unknown;
 const literalString = undefined;
 const literalTrue = undefined;
 const literalNumber = Math.random() > 0.5 ? undefined : undefined;
-const almostPi = undefined;
 const literalObject = {};
 const literalFunction = () => {};


bautistaaa (approved):

challenges/literal-types/user.ts
```diff
@@ -7,5 +7,6 @@ type LiteralFunction = unknown;
 const literalString = undefined;
 const literalTrue = undefined;
 const literalNumber = Math.random() > 0.5 ? undefined : undefined;
+const almostPi = undefined;
```
ishtiaq156, line 10: I think, we don't need almostPi declared. TypeHero challenges expects the user read the tests and figure out what needs to be created. "[Type Aliases](https://typehero.dev/challenge/type-aliases)" challenge has a note regarding this as well.

> As you continue your journey with TypeScript (and TypeHero!) you should get into the habit of closely consulting tests as you go. The tests can help you ensure that your implementation is correct, and quite often that means filling in missing pieces in your implementation.

bautistaaa: shit you’re right, can someone undo that

romain-trotard: Ah yes, well seen. I've just reverted it ;)

PR #1235, opened 2023-12-02
Update countdown-timer.tsx

odicho:
Updated the countdown timer component to auto reset to next midnight everyday!


WVAviator:
#1236 adds dynamic subheader text displaying the date in addition to the countdown timer

Issue #1232, opened 2023-12-01
[Feature]: A public API with user stats


polooner:
### Description

Create a public facing API with user stats. I would love to be able to grab my stats (such as number of problems I solved, maybe % of all challenges) and put it on my portfolio website.

bautistaaa:
i think eventually we will have this and believe this is a great idea. will revisit after the dust settles from AOT

oahnh:
+1 ^^

Dun-sin:
+2 ^O^

matthijsgroen:
+1 This would also allow to have a 'company team' dashboard made, to compare progress on the challenges within our team :-)

PR #1237, opened 2023-12-02
feat:  public api route for getting userstats 

me-imfhd:
Added a basic next get route, with folder named ad userStats, it uses an username query, does prisma call to check if username exists then adds uses the getSolvedChallenges logic to return desired data
## Related Issue

https://github.com/typehero/typehero/issues/1232
Closes #


bautistaaa:
i don’t plan on supporting public apis atm. if they need that information they can see it all on their profile

PR #1249, opened 2023-12-03
disable transition on theme change

staranbeer:
### What happened?

On theme change, some of the components have a visible blinking effect. 

**Examples:**  

Screencast from 03-12-23 03_53_52 PM IST

Screencast from 03-12-23 03_54_18 PM IST

Screencast from 03-12-23 04_18_30 PM IST

### Effected Components
This will effect all of the components that have either `transtion-property : color, background-color` or `transition-property: all`  and  a `transtion-duration`. 

### Solution: 

This can be solved by setting the  `disableTransitionOnChange`  prop to `true` in `next-themes` provider.


zaCKoZAck0:
LGTM, thank you.

Issue #1220, opened 2023-12-01
[Feature]: A "Skip to editor" link for keyboard users


andrewmcoupe:
### Description

When I enter the page for a challenge, keyboard users have to tab at least 10 times before they are focused on the code editor.

Let's have a "skip to code editor" link for keyboard users.

### Explanation

This is useful for keyboard users as they won't have to tab 10+ times to get to the editor to complete their challenge.

junsupark94:
We could put it somewhere in the red region I highlighted.
![Example](https://github.com/typehero/typehero/assets/97586064/0d50085f-7ae0-446b-853c-b500098a50f5)

Upon opening a challenge or refreshing the page, the focus starts at upper left with the TypeHero logo. It goes through the nav items, dark mode, profile, description, something else, then the options button. I was thinking of putting it before the options button.

Also tabbing has to go through comments before it reaches the code editor, even if the comments are hidden. More comments will mean more tabbing. Comment should be not be able to tabbed through when it's hidden.

I also tried navigating with VoiceOver. There's no way to jump to the code editor section. I also have to navigate through everything that tabbing goes through.

I also think the code editor should be focused from the start.

### Summary
1. Add a button to jump to code editor
2. Reorganize the page so screen readers can easily go to code editor
3. Code editor is focused from the start
4. If comments are hidden, it should not be able to be accessed with keyboard navigation or screen readers.

andrewmcoupe:
It should be an anchor element 👍 not a button.

What about linking to the outer wrapper as a first pass? This would require pressing enter on the "Skip to editor" link and then tabbing once more. See my codepen below...

https://codepen.io/andrewmcoupe/pen/dyaQbWx

me-imfhd:
my suggestion: We can use shortcut a shortcut keys, smth like ctrl+shift+J to jump between editor and left panel, vscode has somthing ctrl+shift+e to switch cursor from file explorer and code editor
[Screencast from 02-12-23 07:03:45 PM IST.webm](https://github.com/typehero/typehero/assets/114667178/54a8f22c-d091-4753-a828-97fea5e23973)

junsupark94:
Changed the button to anchor. And I changed the section element that wraps around the first CodeEditor.
`<section id="code-editor" tabIndex={-1} className="h-full overflow-hidden focus:border-blue-500 focus:border">`
[...]

PR #1251, opened 2023-12-03
1220 skip to code editor

junsupark94:
## Description
For Feature https://github.com/typehero/typehero/issues/1220
Adds a "skip to code editor" link in Navigation bar for keyboard users. 

- Readable to screen readers. 
- Only visible when focused. 
- Renders only on Challenge routes.
- Added visual feedback that navigation to code editor has occurred. 

Skip to code editor
Visual feedback


apps/web/src/components/Navigation/skip-to-code-editor.tsx
```diff
@@ -0,0 +1,16 @@
+'use client';
+import { usePathname } from 'next/navigation';
+
+export const SkipToCodeEditor = () => {
+  const pathname = usePathname();
+
+  if (!pathname.includes('challenge')) {
+    return null;
+  }
```
andrewmcoupe, line 9: Maybe checking the presence of a an element with id equal to `#code-editor` would be better?

Or go one step further and make this into a reusable `<SkipTo />` component.

junsupark94: Used document.getElementById to check target element exists. Can't figure out how to defer the check until the page is fully loaded since Navigation loads first and the CodeEditor way later. I tried the following in useEffect:

- setTimeout with 0 delay
- adding event listener on window load
- setTimeout inside the event listener

I don't know how to enable a component to check if another component/element exists elegantly without something like a zustand store.

bautistaaa: use a useEffect / disable SSR for this specific component

junsupark94: > use a useEffect / disable SSR for this specific component

```
  const pathname = usePathname();
  const [shouldShow, setShouldShow] = useState(false);

  useEffect(() => {
    if (pathname.includes('challenge')) {
      setShouldShow(true);
    }
  }, [pathname]);

  if (!shouldShow) {
    return null;
  }
```

Like this?

bautistaaa: oh nevermind misread your original comment

bautistaaa:
this feels like a good step in the right direction. can always improve later

PR #1253, opened 2023-12-04
fix: disable submit buttons when submitting

grumd:
Visual feedback on clicking buttons is important. Buttons that don't react to a click tend to confuse users and get double clicked. Poof poof one-two combo

## Description
Just using `form.formState.isSubmitting` to disable the submit button in all react-hook-form forms that I could find.

## How Has This Been Tested?
Clickety click on a submit button. Typescript seems to be happy.

EQv5BXLmsD


bautistaaa (approved):

bautistaaa:
i thought we did this already lul

PR #1262, opened 2023-12-04
feat: per-challenge tsconfig support

### LINKED ISSUE #1260: [Bug]: the "Assert Array Index" type challenge requires a custom tsconfig
### What happened?

The ["Assert Array Index"](https://github.com/type-challenges/type-challenges/tree/main/questions/00925-extreme-assert-array-index) type challenge requires `[noUncheckedIndexedAccess](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) to be set in the tsconfig to work.

Currently, there's no way to set a custom tsconfig per challenge, but it's reasonably simple to implement and will allow this issue to be resolved cleanly (without clobbering the default config with options that are irrelevant for 99.9% of challenges).

It looks like all "custom" typehero challenges have a custom tsconfig, so implementing this would make it possible to ingest those as well. This would ensure that the tsconfig on the website and the one used for validation are identical, eliminating the possibility of config-related errors.

All

### Relevant log output

```zsh
N/A
```

invakid404:
## Description

This PR adds basic support for per-challenge tsconfigs. TypeHero challenge tsconfigs are now
ingested and applied in the editor. A tsconfig override is set for the "Assert Array Index" type
challenge, which requires `noUncheckedIndexedAccess` to be set.

As explained in [this commit's description](https://github.com/typehero/typehero/pull/1262/commits/deadf459b4acd8f2f874826f7739901f46d5a14e), while it is possible to fetch the tsconfig required for
each of the type challenges, I don't think it makes sense, as "Assert Array Index" is the only type
challenge to require a custom config as of today. I found that fetching /{id}/play for every type
challenge makes ingesting the challenges very slow. I don't think there's enough benefits to justify
the slowdown, so I've opted for hardcoding the tsconfig for the specific challenge that's causing
issues.

## Related Issue

Closes #1260

## Motivation and Context

As explained in #1260, this change allows tsconfig overrides per challenge, which is necessary for
some specific challenges.

## How Has This Been Tested?

I've confirmed that this PR fixes the "Assert Array Index" type challenge, and that it doesn't
appear to break existing challenges. I've also confirmed that the seed scripts and tsconfig
ingestion work correctly.

Screenshot of the "Assert Array Index" tests passing:

image


bautistaaa:
@invakid404 is this ready for review/merge

invakid404:
> @invakid404 is this ready for review/merge

@bautistaaa should be, I've just been rebasing it periodically to keep it up-to-date with main.

packages/db/mocks/challenges.mock.ts
```diff
@@ -18,6 +18,12 @@ export interface InfoFile {
 
 const redundantChallenges = ['pick', 'flatten'];
 
+const tsconfigOverrides: Record<string, object> = {
+  'assert-array-index': {
```
bautistaaa, line 22: isnt this pulled from the tsconfig ingest?

bautistaaa: oh wait does ts challenges not do that

invakid404: As mentioned in the PR description, technically there is a way to automatically ingest the tsconfigs for all type challenges by fetching the "play" URL for each and every one of them, then parsing them from the TypeScript playground URL they redirect you to, but I know for a fact that this is the only challenge that requires a custom tsconfig as of today.

I tried fetching every type challenge's playground URL, but it was awfully slow, so I don't think there's a point in doing this right now.

bautistaaa: that was my bad i just looked at the code and not the description 😂

bautistaaa (approved):

Issue #1244, opened 2023-12-03
[Bug]: Two slash plugin don't work correctly


duckhoa-uit:
### What happened?

The two-slash plugin in the Monaco editor doesn't work correctly. When visiting different challenges, it shows types in a random line.

It requires refreshing to work correctly.

**Reproduce**
1. Visit the day 1 challenge
2. Submit Answer
3. Visit day 2 
4. The two-slash in the editor shows the result in the random line

https://github.com/typehero/typehero/assets/54369632/6975c7ef-ccc3-4300-8048-3f9403330d80

### What browsers are you seeing the problem on?

Chrome

WVAviator:
Yes exactly.

As far as getting it to completely refresh and not show the incorrect hints at all, I've been researching and trying things to no avail. I can continue to hunt for a fix on that if you want.

bautistaaa:
i think this is good enough for now

PR #1265, opened 2023-12-05
Mostly fixed twoslash inlay hints leaking over into other challenges

WVAviator:
## Description

Because InlayHintProviders are registered on the global Monaco instance, it was not being disposed after the user navigates away from the page. Once the user loads up another page with a SplitEditor component, the previous InlayHintsProvider is no longer providing updates.

By wrapping the provided IDisposable from the registerInlayHintProviders function with a ref, it can later be referenced in a useEffect cleanup for the component to call the .dispose() method. This _mostly_ works - however the inlay hints from the previous editor do briefly show on the page for about 200ms. I will continue to search for a fix for this.

I'm opening the PR anyway because this makes twoslash actually usable again for anyone who doesn't know to click refresh.

## Related Issue

Reference #1244 
I don't think it should be closed yet - still working on a more comprehensive fix that fully cleans up the hints between editor instances.

## Motivation and Context

This makes twoslash usable again for someone solving more than one challenge in the same session.

## How Has This Been Tested?

I forked my own version of the type-challenges repo, changed the git repository url in challenges.mock.ts, and then made changes to the challenge test cases that included twoslash comments. I then reseeded the db, and fully reproduced the issue in a test environment. After applying my fix, I observed the issue mostly resolved.

Will provide if desired.


WVAviator:
https://github.com/typehero/typehero/assets/11326285/17ccc526-5f38-4f59-8c0f-471ac5b50003


https://github.com/typehero/typehero/assets/11326285/b7812bbe-1275-4eb7-b415-bb67bbdca7a6

bautistaaa:
if you navigate to another one that has twoslash on it will it render properly?

WVAviator:
> if you navigate to another one that has twoslash on it will it render properly?

Yep

https://github.com/typehero/typehero/assets/11326285/1123a0ec-725c-4071-9b90-6ce5a8d32ae2

bautistaaa (approved):
good enough ™

PR #1268, opened 2023-12-05
feat: #1267 Top Loader Progressbar between page transitions

### LINKED ISSUE #1267: [Feature]: Showing Top load bar between page transitions, for showing page loading progress
### Description

- While using TypeHero WebApp sometimes initially, when switching between pages, it takes more than 1 or 2 seconds so user can be in doubt if the click worked or not. 
- This feature introduces a simple top load bar that appears during page transitions, enhancing the user experience by visually indicating progress. 
- The color and style of the bar will be customizable but will have a default setting that matches the application's design aesthetics.

image

### Explanation

I've used this in other live projects (ayedot.com for example), and it works really good for user experience.

SRX9:
## Description
- Adding a top loading bar using the nextjs-toploader library, with custom CSS styling. 
- This feature will help improve user experience in the TypeHero WebApp by providing a visual page loading progress, during page transitions. 
- It addresses the issue where users may be uncertain if their click was registered due to a delay in loading new pages. 

## Related Issue
Closes #1267 

## Motivation and Context
Since i started using TypeHero, sometimes initially when switching between pages, it takes more than 1 or 2 sometime even 3 seconds to load new page, (mostly during dynamic pages), so in this case it feels like nothing is happening on page for few seconds, so user can be in doubt if the click actually worked or not. 

So, providing a Progress Bar on Top of the page like we see in many popular webapps, would show  users that click is registered and page is being loaded.

## How Has This Been Tested?
The implementation was tested across different browsers (except Safari as i don't use Mac) to ensure compatibility. 

image

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -54,6 +54,7 @@
     "next": "14.0.1",
     "next-international": "1.1.4",
     "next-themes": "^0.2.1",
+    "nextjs-toploader": "^1.6.4",
     "react": "^18.2.0",
     "react-day-picker": "^8.8.0",
     "react-dnd": "^16.0.1",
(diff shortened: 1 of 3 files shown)


bautistaaa:
closing, see comment on original issue

SRX9:
Closing, after discussion in #1267

Issue #1270, opened 2023-12-06
[Bug]: Tooltip not visible on badge hover


zaCKoZAck0:
#### Expected: Tooltip should be visible containing badge name.
Currently it doesn't show anything on hover.
![image](https://github.com/typehero/typehero/assets/69889382/b51a327e-1efe-4235-945e-54b2c9a49fae)

me-imfhd:
can i work on this issue

PR #1271, opened 2023-12-06
fix: on hover badges tooltip fix

me-imfhd:
looking at other tooltip every icon is wrapped with button element, 
badges tooltip icon didn't had any wrapper, thus adding a span wrapper does the work,
if wanted i can do button wrapper instead to span but it will add cursor pointer to it, let me know if we need it

## Related Issue

https://github.com/typehero/typehero/issues/1270
Closes #

[Screencast from 06-12-23 03:57:43 PM IST.webm](https://github.com/typehero/typehero/assets/114667178/90164f04-a3d3-440a-a922-390cb228b858)


zaCKoZAck0:
LGTM Thank you.

PR #1273, opened 2023-12-06
fix(monaco): handle diagnostic message chains

### LINKED ISSUE #1272: [Bug]: incorrect handling of ts diagnostic messages in split editor
### What happened?

Currently, the split editor type checking implementation assumes that all diagnostic messages are strings, but as the types also suggest, some diagnostic messages consist of multiple parts, which results in these errors showing up as `[object Object]` in Monaco.

This is affecting Day 5 of Advent of TypeScript, unintendedly making it harder, due to the fact people can't see the error about readonly types:
image

All

### Relevant log output

```zsh
N/A
```

invakid404:
## Description

This PR fixes the handling of diagnostic message chains in the Monaco split editor. Instead of
assuming that `messageText` is a string, we now call out to TypeScript to flatten the message for
us, which fixes the occasional `[object Object]` error messages.

## Related Issue

Closes #1272.

## Motivation and Context

This PR ensures that error messages are displayed properly in Monaco in order to provide adequate
information about what's wrong to the user.

## How Has This Been Tested?

I've confirmed that the `readonly` error is no longer an issue.

image

### DIFF
--- a/packages/monaco/src/split-editor.tsx
+++ b/packages/monaco/src/split-editor.tsx
@@ -484,7 +484,7 @@ async function typeCheck(monaco: typeof monacoType) {
         endLineNumber: end.lineNumber,
         startColumn: start.column,
         endColumn: end.column,
-        message: d.messageText as string,
+        message: ts.flattenDiagnosticMessageText(d.messageText, '\n'),
       } satisfies monacoType.editor.IMarkerData;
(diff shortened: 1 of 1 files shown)


bautistaaa (approved):
ohh they have a helper, nice!

Issue #1280, opened 2023-12-08
[Bug]: When typing in 256 characters within the Profile bio, both the Challenges and Bio section extend out.


MichaelG824:
### What happened?

Expected: The Challenge/and bio section would stay at a fixed width percentage according to the page. 

<img width="1310" alt="Screenshot 2023-12-07 at 8 57 22 PM" src="https://github.com/typehero/typehero/assets/24363724/99be41cf-24a3-43c5-bda0-5a67bdb6e92f">

Actual: When typing in 256 characters into the Bio, both the Challenges and Bio section extend far out. 

### What browsers are you seeing the problem on?

Google Chrome

zaCKoZAck0:
LGTM, I think we need to break-word on overflow.

MichaelG824:
<img width="866" alt="Screenshot 2023-12-11 at 5 31 08 PM" src="https://github.com/typehero/typehero/assets/24363724/dda9cc5c-32e2-4589-8bed-ab0503327620">
<img width="528" alt="Screenshot 2023-12-11 at 5 31 43 PM" src="https://github.com/typehero/typehero/assets/24363724/82b5e6ae-0dcf-44d9-99de-1eeba4857160">

Looks good on Prod. I'm going to close this issue.

PR #1286, opened 2023-12-09
Issue 1280 bio section extend out

MichaelG824:
## Description
Bio extends out when typing in a really long character string. Hence, max width and break words is set to prevent this type of case from happening. 

## Related Issue
https://github.com/typehero/typehero/issues/1280 

## Motivation and Context
See description

## How Has This Been Tested?
Tested locally on dev with with 256 character and other long strings. 

Screenshot 2023-12-08 at 10 11 45 PM


MichaelG824:
closing to make sure that mobile works too.

PR #1289, opened 2023-12-09
issue-1280 extend out issue on bio with really long character string. 

MichaelG824:
## Description
Bio extends out when typing in a really long character string. Hence, max width and break words is set to prevent this type of case from happening. If there's any disagreement with the max-width just let me know and I can change it! 

## Related Issue
https://github.com/typehero/typehero/issues/1280 

## Motivation and Context
See description

## How Has This Been Tested?
Tested locally on dev on both mobile view (Iphone 14) and browser with strings that have 256 characters and also tested with other long strings. 

Screenshot 2023-12-09 at 4 45 24 PM

Screenshot 2023-12-09 at 4 44 43 PM

### DIFF
--- a/apps/web/src/app/[locale]/(profile)/[username]/_components/dashboard/overview-tab.tsx
+++ b/apps/web/src/app/[locale]/(profile)/[username]/_components/dashboard/overview-tab.tsx
@@ -20,12 +20,12 @@ export async function OverviewTab({ user }: Props) {
         {/* //Todo: Filling with void for now, may put contributions / something else */}
         <div className="max-w-sm flex-grow" />
       </div>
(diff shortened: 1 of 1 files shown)


bautistaaa (approved):

MichaelG824:
hm. It's saying Prettier failed in the CI/CD pipeline but I did run `pnpm run format` and it's stating that `All matched files use Prettier code style!` I'll try to find what else I possibly missed.

bautistaaa:
it should show what’s messed up in the logs, try running pnpm checks:fix

MichaelG824:
Looks like the tailwind css classes were rearranged after running the command. Local testing looks good.

Issue #1287, opened 2023-12-09
[Bug]: Profile, In-Progress, and Shared Solutions within Profile page shows 'Page 1 of 0' on page with No Results. 


MichaelG824:
### What happened?

![Screenshot 2023-12-08 at 10 20 40 PM](https://github.com/typehero/typehero/assets/24363724/7da2147e-4b78-4d91-b8bb-cee8e6ec42e1)

Expected: 'Page 1 of 1'
Actual: 'Page 1 of 0'

### What browsers are you seeing the problem on?

Chrome

MichaelG824:
I can take this as well if that's ok.

MichaelG824:
<img width="951" alt="Screenshot 2023-12-11 at 5 52 21 PM" src="https://github.com/typehero/typehero/assets/24363724/33e75ced-04cc-4bae-bd75-d242d16c4378">
Looks good on Prod

PR #1290, opened 2023-12-10
Issue 1287 - Profile, In-Progress, and Shared Solutions within Profile page shows 'Page 1 of 0' on page with No Results

MichaelG824:
## Description
The PageCount returns 0 on empty set of rows. We should return 1 instead as it would allow us to get Page 1 of 1 instead of Page 1 of 0. 

Edit: We will remove the pagination row instead on 'No Results' Screenshot updated to show what it will look like.  

## Related Issue
https://github.com/typehero/typehero/issues/1287

## Motivation and Context
See description

## How Has This Been Tested?
Tested on local main: 
- number of rows > rows per page and got correct output. 
- number of rows less than number of rows per page and got correct output.
- No number of rows and got correct output.

Screenshot 2023-12-10 at 5 46 55 PM

Screenshot 2023-12-09 at 7 14 47 PM

Screenshot 2023-12-09 at 7 14 57 PM


bautistaaa:
should it just hide then entire pagination row if no results? then you wouldnt need the ternary perhaps

MichaelG824:
@bautistaaa  Yeah I can just remove the row with:  'Rows per Page', Page 1 of 1, etc.  on No Results. That makes sense to me.

bautistaaa (approved):

PR #1292, opened 2023-12-10
chore: Reposition theme button

odicho:
Before:
Light:
image

Dark:
image

After:
Light:
image

Dark:
image

### DIFF
--- a/apps/web/src/components/Navigation/theme-button.tsx
+++ b/apps/web/src/components/Navigation/theme-button.tsx
@@ -1,41 +1,33 @@
 'use client';
-import { useEffect, useState } from 'react';
 import { useTheme } from 'next-themes';
 import { Moon, Sun } from '@repo/ui/icons';
 
 export function ThemeButton() {
-  const [mounted, setMounted] = useState(false);
   const { setTheme, resolvedTheme } = useTheme();
-
-  useEffect(() => {
-    setMounted(true);
-  }, []);
 
   return (
     <>
-      {mounted ? (
-        <button
-          aria-label="theme button"
-          className="group rounded-lg p-2 focus:outline-[#2563EB]"
-          onClick={() => {
-            setTheme(resolvedTheme === 'dark' ? 'light' : 'dark');
-          }}
-          type="button"
-        >
-          {resolvedTheme === 'dark' && (
-            <Moon
-              aria-hidden="true"
-              className="h-5 w-5 duration-150 group-hover:scale-110 group-hover:fill-black dark:group-hover:fill-white"
-            />
-          )}
-          {resolvedTheme === 'light' && (
-            <Sun
-              aria-hidden="true"
-              className="h-5 w-5 duration-150 group-hover:scale-110 group-hover:fill-black dark:group-hover:fill-white"
-            />
-          )}
-        </button>
-      ) : null}
+      <button
+        aria-label="theme button"
+        className="group rounded-lg p-2 focus:outline-[#2563EB]"
+        onClick={() => {
+          setTheme(resolvedTheme === 'dark' ? 'light' : 'dark');
+        }}
+        type="button"
+      >
+        {resolvedTheme === 'dark' && (
+          <Moon
+            aria-hidden="true"
+            className="h-5 w-5 duration-150 group-hover:scale-110 group-hover:fill-black dark:group-hover:fill-white"
+          />
+        )}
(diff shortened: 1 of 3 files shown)


apps/web/src/components/Navigation/theme-button.tsx
```diff
@@ -1,31 +1,33 @@
...
 import { useTheme } from 'next-themes';
 import { Moon, Sun } from '@repo/ui/icons';
 
 export function ThemeButton() {
-  const [mounted, setMounted] = useState(false);
   const { setTheme, resolvedTheme } = useTheme();
 
-  useEffect(() => {
-    setMounted(true);
-  }, []);
```
odicho, line 12: @zachlankton  I noticed you implemented this component and had the mounted state in it. I couldn't really figure out what it is for / what is exactly it does. From my understanding, the first render will return `null` and then the useEffect will run and update the state and trigger a re-render where the button will now render. What is the purpose of doing this? 

ref: https://github.com/zachlankton/typehero/blob/4e136b9316edfe66c6e034f10b539bb10bf57e80/apps/web/src/components/Navigation/theme-button.tsx

odicho: Does the useTheme() hook take some time to initialize maybe? If so, would it make more sense to do something like this?:

```typescript
export function ThemeButton() {
  const { setTheme, resolvedTheme } = useTheme();

  return (
    <>
      {resolvedTheme ?? (
        <button
          aria-label="theme button"
          className="group rounded-lg p-2 focus:outline-[#2563EB]"
          onClick={() => {
            setTheme(resolvedTheme === 'dark' ? 'light' : 'dark');
          }}
          type="button"
        >
          {resolvedTheme === 'dark' && (
            <Moon
              aria-hidden="true"
              className="h-5 w-5 duration-150 group-hover:scale-110 group-hover:fill-black dark:group-hover:fill-white"
            />
          )}
          {resolvedTheme === 'light' && (
            <Sun
              aria-hidden="true"
              className="h-5 w-5 duration-150 group-hover:scale-110 group-hover:fill-black dark:group-hover:fill-white"
            />
          )}
        </button>
      )}
    </>
  );
}
```

odicho: Ahhh, now I see why it is like that, it's for the hydration mismatch. I reverted all the changes regarding this! Sorry about that!

bautistaaa:
im okay with this but it does blend in with the social media icons and can feel "hidden"

bautistaaa (approved):

odicho:
> im okay with this but it does blend in with the social media icons and can feel "hidden"

Yeah I know, I was thinking of receiving some feedback on that actually. I was thinking of perhaps adding some spacing, or even do like other websites: add a section for three icons. Light, Dark and System. Unsure what's ideal here.

bautistaaa:
> > im okay with this but it does blend in with the social media icons and can feel "hidden"
> 
> Yeah I know, I was thinking of receiving some feedback on that actually. I was thinking of perhaps adding some spacing, or even do like other websites: add a section for three icons. Light, Dark and System. Unsure what's ideal here.
> 
> On the other hand, do you think someone looking for this icon will be able to distinguish it from the social media icons?

you could maybe put it on the left and add some space or a | thing with some spacing

or keep it in nav and put in dropdown on mobile(last resort)

odicho:
> > > im okay with this but it does blend in with the social media icons and can feel "hidden"
> > 
> > 
> > Yeah I know, I was thinking of receiving some feedback on that actually. I was thinking of perhaps adding some spacing, or even do like other websites: add a section for three icons. Light, Dark and System. Unsure what's ideal here.
> > On the other hand, do you think someone looking for this icon will be able to distinguish it from the social media icons?
> 
> you could maybe put it on the left and add some space or a | thing with some spacing
> 
> or keep it in nav and put in dropdown on mobile(last resort)

I think pipe character is the move:

image

or

image

apps/web/src/components/Navigation/theme-button.tsx
zachlankton: Lol, you are seeing my refactor.  This change was introduced [here](https://github.com/typehero/typehero/commit/c265875fd494f2d74a58cf60cad8f6631e0f7f43#diff-d9fcccb82e22d45c50029699d79f97851782bbf5a470a6802036498af5158394) by someone named Chris @bautistaaa ?? ;-)

Issue #1281, opened 2023-12-08
[Feature]: Update to suggested prisma adapter package from auth.js docs 


harshmangalam:
### Description

- pnpm remove @next-auth/prisma-adapter
+ pnpm i @auth/prisma-adapter

PR #1282, opened 2023-12-08
[Feature]: Update to suggested prisma adapter package from auth.js docs

harshmangalam:
#1281

PR #1302, opened 2023-12-10
prisma-adapter updated (#1281)

hatsmaker:
## Description

Removed the `@next-auth/prisma-adapter` and  after installed `@auth/prisma-adapter` as described in the issue's description. Changed the `PrismaAdapter` import from `@next-auth/prisma-adapter` to `@auth/prisma-adapter` in `packages/auth/src/server.ts`
## Related Issue

Closes https://github.com/typehero/typehero/issues/1281

### DIFF
--- a/packages/auth/src/server.ts
+++ b/packages/auth/src/server.ts
@@ -1,4 +1,4 @@
-import { PrismaAdapter } from '@next-auth/prisma-adapter';
+import { PrismaAdapter } from '@auth/prisma-adapter';
 import type { Role, RoleTypes } from '@repo/db/types';
 import { type DefaultSession } from 'next-auth';
 import GitHubProvider from 'next-auth/providers/github';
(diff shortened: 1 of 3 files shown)


bautistaaa:
this is a duplicate of https://github.com/typehero/typehero/pull/1282

theres type errors though so whoever resolves first wins :)

packages/auth/src/server.ts
```diff
@@ -104,7 +105,7 @@ export const {
       };
     },
   },
-  adapter: PrismaAdapter(prisma),
+  adapter: PrismaAdapter(prisma) as Adapter,
```
bautistaaa, line 108: feels weird we have to cast for official adapters. is this from the docs?

hatsmaker: There is a weird type mismatch error that occurs without an explicit cast, it seems there are some incompatibilities between `next-auth` and `@auth/prisma-adapter` Adapter types.
Link: https://github.com/nextauthjs/next-auth/issues/7727

bautistaaa: this issue seems to suggest the type error is supposed to be resolved in 5. oh well i’ll look deeper later

bautistaaa (approved):

PR #1331, opened 2023-12-16
fix(search): don't perform empty queries

invakid404:
## Description

This PR fixes the infamous Instantsearch query on mount issue by detecting empty searches and
resolving with an empty response. Previously, every time someone loaded the landing page (or any
page that has the search box), a query was performed, which racked up a lot of requests. Not ideal,
considering there's a quota.

## Related Issue

N/A

## Motivation and Context

Basically the description.

## How Has This Been Tested?

I've confirmed locally that the newly added code correctly detects empty queries and handles them
locally as opposed to hitting Algolia.

N/A


bautistaaa (approved):
GOAT

PR #1342, opened 2023-12-22
fix: remove `z-index` from header

invakid404:
## Description

This PR removes the `z-index` from the header, which is the cause of elements in the header clipping
through tooltips that show over it.

## Related Issue

N/A

## Motivation and Context

This was an issue on the challenge page, as the tooltips of the buttons that are at the top of the
code editor show up over the header.

## How Has This Been Tested?

I've confirmed that all the links and buttons in the header appears to work correctly without a
`z-index`, so it should be fine.

Before the change:

image

After the change:

image

### DIFF
--- a/apps/web/src/app/[locale]/_components/hero.tsx
+++ b/apps/web/src/app/[locale]/_components/hero.tsx
@@ -147,11 +147,11 @@ export async function Hero() {
   const t = await getScopedI18n('landing.hero');
   const session = await auth();
   return (
-    <section className="-mt-[56px] min-h-[calc(100vh)] overflow-hidden lg:min-h-0 lg:pt-[56px]">
+    <section className="pointer-events-none -mt-[56px] min-h-[calc(100vh)] overflow-hidden lg:min-h-0 lg:pt-[56px]">
       <div className="absolute inset-10 -z-30 overflow-hidden rounded-full opacity-70 lg:hidden">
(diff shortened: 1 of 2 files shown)


invakid404:
Unfortunately, this is broken on the landing page for some reason. I'll investigate.

invakid404:
The issue is that the hero section has a negative margin and overlaps with the header (which is probably why the z-index was useful in the first place). Nevertheless, I managed to fix it without z-index by disabling pointer events on the parent container (the one that overlaps) and re-enabling them for all children, which appears to work just fine.

bautistaaa:
this disabled being able to click on anything in the nav on the homepage. Its probably because pointer event none on the hero component overlaps header me thinks

invakid404:
@bautistaaa did you redeploy the preview/test this locally? The pointer-events-none is supposed to fix that.

bautistaaa:
> @bautistaaa did you redeploy the preview/test this locally? The pointer-events-none is supposed to fix that.

thought I did. ill redeploy now

invakid404:
@bautistaaa looks like it was still broken on mobile, which I somehow missed, but _now_ it should work lol

bautistaaa:
this appears to work. fire in the hole

bautistaaa:
looks like its broken on 404 page >.<

PR #1345, opened 2023-12-23
Show Avent Of Code cards as links once they are flipped

### LINKED ISSUE #1344: [Feature]: Advent of code challenge card should show a URL
### Description

At the moment, Advent of code cards, once flipped, are not displayed as links.

The fix should be simple.

Incomming PR

### Explanation

Not showing a URL on something that redirects the user is an accessibility issue.
We should care about those disabled users.

ecyrbe:
## Description

For context: see this tweet :
https://twitter.com/HugeLeters/status/1738596106034946139
## Related Issue

Closes #1344
## Motivation and Context

It's an accessibility issue.
## How Has This Been Tested?

I tested it on local dev, i took care to prevent default so that the page don't fully reloads, but uses the router.

### DIFF
--- a/apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx
+++ b/apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx
@@ -197,10 +197,13 @@ export function TiltableCard({ index, challenge }: Props) {
           </p>
         </motion.div>
       </motion.div>
-      <motion.div
+      <motion.a
+        href={`/challenge/${challenge.slug}`}
         onClick={(e) => {
           e.stopPropagation();
+          e.preventDefault();
           router.push(`/challenge/${challenge.slug}`);
+          return false;
         }}
         initial={{ rotateY: 180 }}
         animate={{ rotateY: isFlipped ? 0 : 180 }}
@@ -268,7 +271,7 @@ export function TiltableCard({ index, challenge }: Props) {
           )}
           <p className="text-xl font-bold text-white">{challenge.name}</p>
         </motion.div>
-      </motion.div>
+      </motion.a>
     </motion.div>
   );
 }


apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx
```diff
@@ -197,10 +197,13 @@ export function TiltableCard({ index, challenge }: Props) {
           </p>
         </motion.div>
       </motion.div>
-      <motion.div
+      <motion.a
+        href={`/challenge/${challenge.slug}`}
```
Comment (bautistaaa) [inline apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx:201]: I think using `Link` would be better but same result I suppose

bautistaaa (approved):

apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx
```diff
@@ -197,10 +197,13 @@ export function TiltableCard({ index, challenge }: Props) {
...
         onClick={(e) => {
           e.stopPropagation();
+          e.preventDefault();
           router.push(`/challenge/${challenge.slug}`);
+          return false;
```
Comment (TkDodo) [inline apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx:206]: If there is an `a` tag with an href now, can't we completely remove the `onClick` handler?

Comment (bautistaaa) [reply apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx:206]: wouldn’t that cause a full page reload since we’re not using the router

Comment (ecyrbe) [reply apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx:206]: yes. and i didn't use Link, because i was not sure what motion is doing here. But i can definitely flip to Link if prefered

Comment (bautistaaa) [reply apps/web/src/app/[locale]/aot-2023/_components/tiltable-card.tsx:206]: this code won’t live much longer so not a big deal

PR #1351, opened 2023-12-24
Add:[simple meaing of primitive dataType]

Ghada-Emad1:
## Description

Another Simple Meaning of Primitive Data Type
- we give every variable or anything that you deal with in Code A simple Type that will reaturn or what variable will contain , if it number so the type of it is number ,if it string i will give it type of string and so on

## Related Issue

Closes #


challenges/primitive-data-types/user.ts
```diff
@@ -8,11 +8,13 @@ const age = 52;
 
 interface Musician {
   artistName: string;
```
bautistaaa, line 10: the whole point of the challenge is for the user to do this themselves

bautistaaa:
prompt is fine as is

PR #1366, opened 2023-12-29
updated svg props to have optional title as well as added title attri…

### LINKED ISSUE #1364: [Bug]: e2e tests are failing for unknown reasons
### What happened?

The a11y test started failing after tweaking some unrelated html

commit: https://github.com/typehero/typehero/commit/e46a4ffb4b9fbe32583222973cca54a8625f2683

_No response_

### Relevant log output

_No response_

mdwiltfong:
…bute to respective svgs fixes #1364

Looking at the failed [gh action workflow](https://github.com/typehero/typehero/actions/runs/7340830844/job/19987500547) it looks like the a11y tests on the home page are triggering this [violation](https://dequeuniversity.com/rules/axe/4.8/svg-img-alt?application=playwright). 

I think one way around solving the violation and removing the `<title>` tags is by modifying the `SVGProps` type to include an option `title` attribute for SVGs. 

## Description

Changes SVGProps to include an optional title attribute

## Related Issue

Closes #1364  

## How Has This Been Tested?

Ran playwright locally and i'm no longer seeing the failure.


bautistaaa:
looks like TS hates this.

i did some brief research and `title` tags in svg's _shouldnt_ impact SEO so we can just revert the title tag changes.
do you mind adding those back in?

mdwiltfong:
> looks like TS hates this.
> 
> i did some brief research and `title` tags in svg's _shouldnt_ impact SEO so we can just revert the title tag changes. do you mind adding those back in?

That's odd because I updated those types. When performing the type check, does the workflow not use the types in the PR? Just curious! 

I added those title tags back, and reverted the type to what it was originally. Going to keep an eye on the tests.

bautistaaa:
> > looks like TS hates this.
> > i did some brief research and `title` tags in svg's _shouldnt_ impact SEO so we can just revert the title tag changes. do you mind adding those back in?
> 
> That's odd because I updated those types. When performing the type check, does the workflow not use the types in the PR? Just curious!
> 
> I added those title tags back, and reverted the type to what it was originally. Going to keep an eye on the tests.

it should be using the same typechecking

mdwiltfong:
Taking a look at these failures 👁️

mdwiltfong:
Okay I think all the a11y violations getting picked buy the axe-core engine are resolved. There were three in total: 

1. SVGs not having some sort of title reference. This was resolved by adding the `<title>` tags back
2. The copy button that was introduced needed either a title attribute or an aria-label attribute. I added an aria-label attribute.
3. The shield(?) in the `hero-challenge-card` did not have enough of a color contrast between it's foreground and the background (only in light mode). As a result I changed the foreground of the badge in order to increase the contrast. I used this [tool](https://dequeuniversity.com/rules/axe/4.8/color-contrast?application=playwright) to confirm that it would pass any WCAG AA violation. Although it will fail AAA. I've attached the modified color below. I've tested this in dark mode, and the change seems fine. 
Screenshot 2023-12-30 at 12 05 28 PM

mdwiltfong:
Nice! All the GH Workflows are passing.

bautistaaa:
lfg

Issue #1210, opened 2023-11-30
[Bug]: Someone else profile


fecony:
### What happened?

/Didn't explore repository code/ 

Just joined on website, opened my account and there is someone else 😭 
https://typehero.dev/@Richard
![Screenshot 2023-11-30 at 16 38 23](https://github.com/typehero/typehero/assets/36774784/45118c21-0c52-4241-b7dd-a6ccee8389e4)

Maybe profile pages based on name and not github username is not best idea.

If it's just a placeholder image/profile, then anyway /@username will be unique and not /@name profile url  :eyes:

### What browsers are you seeing the problem on?

All

WVAviator:
I was also just thinking that it'd be nice if I could change my username. I'd rather use my handle @WVAviator than my full name. There does not appear to be an option to do so currently in settings.

bautistaaa:
hmm we use your login, which should be unique `profile.login`

bautistaaa:
> I was also just thinking that it'd be nice if I could change my username. I'd rather use my handle @WVAviator than my full name. There does not appear to be an option to do so currently in settings.

whats your profile url?

bautistaaa:
working on fix @fecony we'll likely need to do some manual stuff for your account and whoever else that is impacted

https://github.com/typehero/typehero/pull/1212

fecony:
> working on fix @fecony we'll likely need to do some manual stuff for your account and whoever else that is impacted
> 
> 
> 
> https://github.com/typehero/typehero/pull/1212
> 
> 

It's has nothing yet so whatever 😄

bautistaaa:
@fecony @WVAviator 

can yall talk in discord so i can try to update your name to your login

bautistaaa:
@fecony 

https://typehero.dev/@fecony

fecony:
@bautistaaa Thank you for fixing this so quickly! ❤️

Hacksore:
@bautistaaa did you commit your fix script anywhere?

ishtiaq156:
> working on fix @fecony we'll likely need to do some manual stuff for your account and whoever else that is impacted
> 
> #1212

looks like I'm impacted as well.
Here's my actual profile link - https://typehero.dev/@Ishtiaq%20Ul%20Haq%20Syed
I would like it to be https://typehero.dev/@ishtiaq156 - which has same handle as GitHub's https://github.com/ishtiaq156

bautistaaa:
[...]

PR #1376, opened 2024-01-01
fix(auth): remove patch of `@auth/core`

juliusmarminge:
Removes the next-auth patch and uses the `profile` option instead to set `p.login`


packages/auth/src/server.ts
```diff
@@ -100,7 +100,7 @@ export const {
           ...session.user,
           id: user.id,
           role: userRoles,
-          createAt: user.createdAt,
+          createdAt: user.createdAt,
```
juliusmarminge, line 103: idk if this was intentionally spelt `createAt`?

bautistaaa: woops

apps/auth-proxy/routes/[...auth].ts
```diff
@@ -11,6 +11,12 @@ export default eventHandler((event) =>
       Github({
         clientId: process.env.GITHUB_ID,
         clientSecret: process.env.GITHUB_SECRET,
+        profile: (p) => ({
```
Comment (juliusmarminge) [inline apps/auth-proxy/routes/[...auth].ts:14]: i don't think this is necessary, since the proxy shouldn't touch this thing anyway but put it in here to be sure

packages/auth/src/server.ts
```diff
@@ -59,57 +59,57 @@ export const {
...
   },
-  adapter: PrismaAdapter(prisma) as Adapter,
   providers: [
     GitHubProvider({
       clientId: process.env.GITHUB_ID!,
       clientSecret: process.env.GITHUB_SECRET!,
+      profile: (p) =>
+        ({
+          id: p.id.toString(),
+          name: p.login,
```
Hacksore, line 109: Is `.login` always the username?

We initially introduced the patch cause some users have display names like me as "Sean Boult" but most don't.

We want @<username> to always be the happy path just want to confirm this before we merge.

Hacksore: related context https://github.com/typehero/typehero/issues/1210

bautistaaa: that’s what the patches does. uses login vs name(which is not unique)

juliusmarminge: All the patch did was change

```diff
- name: p.name ?? p.login
+ name: p.login
```

which is what this does too

bautistaaa:
tested a bit locally and seems fine to me

@juliusmarminge are the methods you're overriding no-ops? is there potentially something else we should be doing(whatever is default) in addition to what you did in the overrides or does it already handle everything?

bautistaaa (approved):

PR #1389, opened 2024-01-05
fix: add the missing almostPi constant in literal-types challenge

jiangshanmeta:
## Related Issue

Closes #

## Motivation and Context

There is a test case about almostPi constant.
But there is no related declaration in the user input area.
That's very confusing.


Brkgng:
@bautistaaa While checking for any open issue related to this challenge, I noticed this PR. How about adding a comment to the place where the 'almostPi' test is located? It seems a bit confusing at the moment.

PR #1391, opened 2024-01-05
chore: clean up types + deps

bautistaaa:
- chore: try to clean up versions
- chore: remove next auth dep

## Related Issue

Closes #


bautistaaa:
HELP!

i cant get these type castings to work

https://github.com/typehero/typehero/blob/trash/misc/packages/auth/src/server.ts#L62-L90

packages/db/schema.prisma
```diff
@@ -100,7 +100,7 @@ model User {
   id String @id @default(uuid())
 
   name          String
-  email         String?   @unique
+  email         String   @unique
```
juliusmarminge, line 103: this should be safe since we always get an email from github

PR #1397, opened 2024-01-10
Add vim type hover focus update

### LINKED ISSUE #802: [Bug]: VIM input breaks when showing type info twice
### What happened?

repro:

1. go to any challenge
2. enter vim mode
3. move your editor cursor on a symbol/var
4. `SHIFT` + `k` twice
5. borked

---

https://github.com/typehero/typehero/assets/996134/b1cce554-153c-447e-9b70-824a1ab8eeb7

_No response_

### Relevant log output

_No response_

heathhenley:
## Description
I know this is a little hacky, def open to other ideas to look into. Basically, I think that if you use the keyboard shortcut to try to open the hover when it's already open, it does toggle to closed as you would expect, but when it's open it's focused, and when it toggles closed it doesn't refocus the editor element, instead nothing is focused. So this just focuses the editor if we're in vim mode and nothing is focused. 

## Related Issue
Closes #802 

## Motivation and Context
There's way more details and context in the issue linked above.

## How Has This Been Tested?
Run locally and tested.

https://github.com/typehero/typehero/assets/5819478/179a1fbe-4e61-4f27-9a6c-b47dc9e4feca

FYI @Hacksore

### DIFF
--- a/packages/monaco/src/split-editor.tsx
+++ b/packages/monaco/src/split-editor.tsx
@@ -95,6 +95,17 @@ export default function SplitEditor({
   const inlayHintsProviderDisposableRef = useRef<monacoType.IDisposable>();
 
   useEffect(() => {
+    const keepEditorFocusedOnHoverCloseVim = (e: KeyboardEvent) => {
+      // In VIM mode, if the type hint hover is open, and we use the keyboard
+      // to close it, it's blured and nothing is focused anymore, so this
+      // is a workaround to keep the editor focused.
+      if (settings.bindings !== 'vim') {
+        return;
+      }
+      if (document.activeElement === document.body) { // nothing is focused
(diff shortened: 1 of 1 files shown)


heathhenley:
> need to think how this impacts keyboard users somehow because this is on any keydown

good point - I tried to add the handler to `onKeyDown` in the editor `section` itself, and that originally worked when I first looked at the issue, (#802) but testing today that approach doesn't seem to work anymore. I guess that whenever handler the action triggers to close the type hover and blur must happen after the handler  on `section`.

heathhenley:
Just going to close to clean up as I'm not sure the right way to fix this and it seems like it's probably pretty low priority

Hacksore:
thanks and I'll close the issues as it's not super important.

PR #1412, opened 2024-02-04
Update admin image view

Hacksore:
Use a flex layout and make the images linkabe.

Error images will not work cause `utfs.io` doesn't return a 404 when images are missing ([discord thread](https://discord.com/channels/966627436387266600/1203707274800013352))


jhechtf (approved):
Code LGTM but I'm apparently not an admin so i can't tell for sure.

PR #1423, opened 2024-02-14
fix: double problem explorer issue

me-imfhd:
When not enrolled from any track, two problem explorer were popping up,
Removed the old boolean statement which used to check whether explorer is disabled or not,
and showed the ui for problem explorer, now it just checks whether the currect problem is a track challenge or not which is null if user is not enrolled in any track automatically
## Related Issue

Closes #


me-imfhd:
image
i made this context for checking whether explorer should be disabled or enabled which used to be determined by this api
image

with the latest code we dont need this, and can be removed, just the thing is in future if we want any means through which you want to disable the problem explorer conditionally, this api can be altered and the context in the first screenshot can be reused, but we don't need this in current scenario, so i don't know really well whats the best. So you can decide.

@bautistaaa 
@jhechtf 

PS: the PR is ready to merge it fixes that issue for good. Sorry for the inconvenience at start.

bautistaaa (approved):

PR #1432, opened 2024-03-09
Add CSP header for better security™

### LINKED ISSUE #1403: [Feature]: Add Content-Security-Policy
### Description

incase we get some moar XSS

https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

### Explanation

keep us safe perhaps?

Hacksore:
I have a few blockers ATM but it seems to "work".

- @d0nutptr take a look at the CSP rules
- @QuiiBz help me figure out if this is ok with i18n

I do seem to get flash banged for some reason with this change though 🤔.


Hacksore:
Someone res this please 🏃‍♂️💨

PR #1440, opened 2024-04-09
docs: Remove PlanetScale recommendation for local setup

jhechtf:
With PlanetScale removing their hobby tier, we shouldn't be pushing people over to their service for a local-only project that they will likely have to pay for.

## Description

Updates the LOCAL.md file to remove the recommendation for PlanetScale in local dev. Now we only have instructions for Docker, with a mention for any MySQL compatible host that works with Prisma.

## Related Issue
N/A

## Motivation and Context
Planetscale removed their hobby tier and we were pushing people to planetscale

## How Has This Been Tested?
N/A

It's an MD file.

### DIFF
--- a/LOCAL.md
+++ b/LOCAL.md
@@ -13,8 +13,7 @@ Provide your values as needed.
 
 ### 2 Configure your database
 
-You can either use PlanetScale or a local Docker container to run your database.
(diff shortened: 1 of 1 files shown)


Hacksore (approved):
:shipit:

PR #1460, opened 2024-07-23
feat: scroll on left side pane

### LINKED ISSUE #1392: [Feature]: Keyboard Arrows Should Scroll the Left Pane
### Description

When the left side pane is focused, in the description tab for instance, when I press the down arrow on my keyboard, I would expect the content to scroll. Currently it does not.

### Explanation

_No response_

peterhangg:
## Description

When the left side pane is focused, users aren't able to scroll down or up using the arrow keys

## Related Issue

Closes #1392 

## Motivation and Context

- Added `keydown` event listener to the description pane when it's focused to be scrollable.
- Div elements require the `tabIndex` attribute to be focusable with the value of `-1` to prevent the element from being navigated directly using the tab key.

### Before
https://github.com/user-attachments/assets/61afa035-e182-4280-927a-14ba9d827b72
### After
https://github.com/user-attachments/assets/73c3d48e-57c6-4f5f-ad94-aff5b30bfba4

### DIFF
--- a/apps/web/src/app/[locale]/challenge/_components/description/index.tsx
+++ b/apps/web/src/app/[locale]/challenge/_components/description/index.tsx
@@ -92,7 +92,7 @@ export function Description({ challenge }: Props) {
     <div
       className="custom-scrollable-element h-full overflow-y-auto px-4 pb-36 pt-3 focus:outline-none"
       ref={descriptionRef}
-      tabIndex={0}
+      tabIndex={-1}
     >
       <div className="flex items-center">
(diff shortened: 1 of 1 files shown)


bautistaaa:
do we really need javascript to scroll with arrow keys? scrolling feels really slow when holding down the arrow keys on long content

bautistaaa:
https://github.com/typehero/typehero/commit/618f04f3ab88b8cb2ac5177e93931f809f1abf35

PR #1461, opened 2024-07-27
chore: remove intl lib

bautistaaa:
## Description
We don't really have the capacity to support translations or even need them. It has caused some weird workaround (404s, static pages not being static). It is best to remove for now and revisit later
## Related Issue

Closes #


bautistaaa:
fire in the hole

PR #1503, opened 2024-08-07
chore: update turbo

bautistaaa:


bautistaaa:
@anthonyshew why it so hard to get turbo 2.0 

me no understand the errors in the quality checks

PR #1505, opened 2024-08-08
add: schema for notifications

bautistaaa:
Does this make sense?

leverage `type` for notification templating (eg: `liked your comment`, `replied to your comment/solution`, `{x} mentioned you`

currently we'll only do notifications comment activity (like/reply) and submitted solutions(like/reply)

### DIFF
--- a/packages/db/schema.prisma
+++ b/packages/db/schema.prisma
@@ -16,6 +16,8 @@ enum NotificationType {
   LIKE
   REPLY
   MENTION
+  MISC
+  SYSTEM
 }
 
 enum Tags {
--- a/packages/db/schema.prisma
+++ b/packages/db/schema.prisma
@@ -436,6 +436,7 @@ model Notification {
   id               Int              @id @default(autoincrement())
   createdAt        DateTime         @default(now())
   updatedAt        DateTime         @updatedAt
+  isRead           Boolean          @default(false)
   type             NotificationType
   toUser           User             @relation("notificationToUser", fields: [toUserId], references: [id])
   toUserId         String
--- a/packages/db/schema.prisma
+++ b/packages/db/schema.prisma
@@ -2,7 +2,7 @@
 // learn more about it in the docs: https://pris.ly/d/prisma-schema
 
 generator client {
-  provider = "prisma-client-js"
+  provider      = "prisma-client-js"
   binaryTargets = ["native", "rhel-openssl-1.0.x"]
 }
 
@@ -10,6 +10,12 @@ datasource db {
   provider     = "mysql"
   url          = env("DATABASE_URL")
   relationMode = "prisma"
+}
+
+enum NotificationType {
+  LIKE
+  REPLY
+  MENTION
 }
 
 enum Tags {
@@ -101,7 +107,7 @@ model User {
   id String @id @default(uuid())
 
   name          String
-  email         String   @unique
+  email         String    @unique
   emailVerified DateTime?
   image         String?
   roles         Role[]
@@ -130,9 +136,11 @@ model User {
   reportsAbout Report[] @relation("reported_user")
   moderated    Report[] @relation("report_moderator")
 
-  imageUpload ImageUpload[]
-  tracks      Track[]
-  BetaTokens  BetaTokens[]
+  imageUpload          ImageUpload[]
+  tracks               Track[]
+  BetaTokens           BetaTokens[]
+  toUserNotification   Notification[] @relation("notificationToUser")
+  fromUserNotification Notification[] @relation("notificationFromUser")
 }
 
 model VerificationToken {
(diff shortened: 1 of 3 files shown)


packages/db/schema.prisma
```diff
@@ -12,6 +12,12 @@ datasource db {
   relationMode = "prisma"
 }
 
+enum NotificationType {
```
zaCKoZAck0, line 15: Can we add a miscellaneous type jic we add some more notifications in the future for something like you earned a badge or any news we want to share through notifications and so on.

enum NotificationType {
  LIKE
  REPLY
  MENTION
  MISC
}

Hacksore: Some other options:

- `SYSTEM` used for critical updates to users
- `NEW_FOLLOWER` if we wanna be social 

and I agree with @zaCKoZAck0 about a `MISC` as a catch all.

packages/db/schema.prisma
```diff
@@ -419,3 +429,25 @@ model BetaTokens {
 
   @@index([userId])
 }
+
+model Notification {
+  id               Int              @id @default(autoincrement())
```
zaCKoZAck0, line 434: Now that I think of it, should we add
`isRead      Boolean`
`/// Redirects to this url on notification click`
`url?       String`
??

packages/db/schema.prisma
bautistaaa: i like SYSTEM. we can add follows at some point but doubt that'll be a thing for a bit

packages/db/schema.prisma
bautistaaa: ah shit thought i added that

bautistaaa: we'll derive the url from the foreign key fields i think?

kfahad5607: I think storing the URL in the table would be useful in situations where the notification type alone is insufficient to generate the URL, such as system notifications, news, external tutorials, blogs, contests, videos, etc. Additionally, it allows for cases where we may want to override the action URL.

packages/db/schema.prisma
```diff
@@ -419,3 +431,25 @@ model BetaTokens {
...
+  createdAt        DateTime         @default(now())
+  updatedAt        DateTime         @updatedAt
+  type             NotificationType
+  toUser           User             @relation("notificationToUser", fields: [toUserId], references: [id])
+  toUserId         String
+  fromUser         User             @relation("notificationFromUser", fields: [fromUserId], references: [id])
+  fromUserId       String
+  sharedSolutionId Int?
```
kfahad5607, line 444: Have we considered deduplication or aggregation of notifications? This would be particularly useful in scenarios where multiple similar notifications occur within a short span of time. For example, when multiple users upvote a solution, like a comment, or reply to a comment, instead of creating a separate entry in the notification table for each action (which could result in an overwhelming number of notifications), we could aggregate or group similar notifications.

packages/db/schema.prisma
zaCKoZAck0: Probably for the specified categories, yes. But we might need it for `MISC` notifications.

zaCKoZAck0: That's why I suggested it to be optional `url?`

packages/db/schema.prisma
zaCKoZAck0: No need ATM, we don't have Taylor Swift posting on our platform 😭

packages/db/schema.prisma
bautistaaa: ok ill add it

packages/db/schema.prisma
bautistaaa: probably something to consider post MVP of this

bautistaaa: i would like to know how we can accomplish that though @kfahad5607

bautistaaa:
we okay pushing this?

packages/db/schema.prisma
```diff
@@ -419,3 +431,26 @@ model BetaTokens {
...
+  sharedSolution   SharedSolution?  @relation(fields: [sharedSolutionId], references: [id])
```
kfahad5607, line 445: Instead of adding multiple resource-specific columns to the table, such as `comment`, `sharedSolution`, etc., could we use a more generic approach with something like `objectType` and `objectId`? This would allow notifications to be associated with any type of object. Given that we maintain data integrity at the application level, this approach would simplify the schema and accommodate various notification types.

bautistaaa: prisma doesn't allow for polymorphic relationships. i'd rather not lose the strong typing tbh

kfahad5607: Understood!

zaCKoZAck0:
> we okay pushing this?

did you add `isRead`, I don't see it. We don't need to track it? just asking.

bautistaaa:
> > we okay pushing this?
> 
> did you add `isRead`, I don't see it. We don't need to track it? just asking.

woops will add in a bit

PR #1506, opened 2024-08-08
fix: Prevent '0' being passed as the page param to paginated solution comments api

### LINKED ISSUE #1504: [Bug]: Deletion of Last Comment Results in Prisma Error During Paginated Comments Fetch
## Description

After deleting the last remaining comment of a solution (**i.e., the final comment left, not the last one in order**), we re-fetch the paginated comments with the `page` parameter value set to 0, which causes a Prisma error on the server.

## Reproduction of Error

1. **Delete the last remaining comment of a solution**

   Delete Comment

2. **Client fetches paginated comments but the request fails**

   Fetch Comments Error

3. **The request fails because the `page` parameter value sent to the server is 0**

   Page Parameter Issue

4. **The `page` parameter being 0 causes the value of the `skip` parameter to be negative, resulting in a Prisma error**

   Prisma Error

   Prisma Error Details

## Probable Solutions

1. Add a check on the client to avoid sending invalid `page` values.
2. Add a check on the server to handle invalid `page` values.
3. Checks on both the client and server.

**PS:** I would be more than happy to work on this issue. Please assign it to me if possible.

All

### Relevant log output

```zsh
web:dev: Error in query graph construction: AssertionError("Invalid value for skip argument: Value can only be positive, found: -10")
web:dev:     at async getPaginatedComments (./src/app/challenge/_components/comments/getCommentRouteData.ts:113:22)
web:dev: digest: "2517259583"
web:dev:  POST /challenge/hello-world/solutions/51 500 in 107ms
web:dev: prisma:error 
web:dev: Invalid `prisma.comment.findMany()` invocation:
```

kfahad5607:
## Description
Added checks to prevent a value of 0 from being passed as the page parameter for the paginated solution comments API.

## Related Issue
Fixes #1504

## Motivation and Context
When deleting the last comment on a solution, the client was fetching paginated comments with a page parameter value of 0. I have added checks on the client to prevent invalid values from being passed to the API call. Additionally, I have added a check on the server to prevent negative values from being passed as the `skip` parameter for Prisma.

## How Has This Been Tested?
Tested by deleting the last comment on a solution, ensuring that the request made to the server contains a page parameter value of 1.

## Screenshots:
Last Comment:
image

Proper page parameter passed:
image

### DIFF
--- a/apps/web/src/app/challenge/_components/comments/comments.hooks.tsx
+++ b/apps/web/src/app/challenge/_components/comments/comments.hooks.tsx
@@ -93,7 +93,10 @@ export function useComments({ type, rootId, initialPage }: UseCommentsProps) {
           description: 'The comment was successfully deleted.',
         });
       }
-      const newPage = data?.comments.length === 1 && commentsMeta.page > 1 ? commentsMeta.page - 1 : commentsMeta.page;
+      const newPage =
+        data?.comments.length === 1 && commentsMeta.page > 1
+          ? commentsMeta.page - 1
+          : commentsMeta.page;
       changePage(newPage);
       queryClient.invalidateQueries({
         queryKey: getQueryKey({ sort: commentsMeta.sort.value, page: newPage }),
--- a/apps/web/src/app/challenge/_components/comments/comments.hooks.tsx
+++ b/apps/web/src/app/challenge/_components/comments/comments.hooks.tsx
@@ -93,7 +93,7 @@ export function useComments({ type, rootId, initialPage }: UseCommentsProps) {
           description: 'The comment was successfully deleted.',
         });
       }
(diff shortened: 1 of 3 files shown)


bautistaaa (approved):

kfahad5607:
@bautistaaa  Do I need to do anything else on this PR?

bautistaaa:
> @bautistaaa  Do I need to do anything else on this PR? 

i approved it already but your CI is failing

bautistaaa:
you need to run pnpm checks:fix or whatever the command is cause prettier is failing

kfahad5607:
Ran the format scripts. I will keep this in mind for future PRs. Thank you.

Issue #1478, opened 2024-08-02
[Feature]: Mentions


bautistaaa:
### Description

Users should be able to tag other users in comments and other applicable areas

related to: https://github.com/typehero/typehero/issues/1477

dca123:
I am trying to understand why ? If there is a need outside of comments I understand (not sure where that is at tho), but wouldn't reply & comment notifications suffice here ? 

I imagine an interaction would go like this

- Solution posted by User A -> User B asks for a clarification as a comment -> User A makes a reply to that comment. 
- Solution posted by User A (he used awaited in the promise challenge) -> User B replies (bruh)

dca123:
Since replies aren't treated as comment, but rather sub comments (I actually dig this implementation since a solution thread won't get bogged down with the conversation of two folks), I'm making an assumption that comments will be mutually exclusive of each other - are there any cases where someone might want to make a tag, but not make a reply ?

Issue #1477, opened 2024-08-02
[Feature]: Notifications


bautistaaa:
### Description

Users should be able to receive notifications when

- someone replies to their comment/solution
- tags them in a comment
- someone likes their comment/solution

We can just poll for the initial version.

kfahad5607:
What kind of notifications we are talking about? Email notifications or in-app notifications or something else? Would love to contribute to this feature.

bautistaaa:
> What kind of notifications we are talking about? Email notifications or in-app notifications or something else? Would love to contribute to this feature.

for the initial version just in app notifications

PR #1508, opened 2024-08-13
feat(mentions): mentions + notifications

bautistaaa:
### Description
Enables the ability to tag any user in comments.
Enables Notifications. See below for behavior

known issue: 
- you can duplicate notifications (e.g: unlike and like a comment will send another notification)
- editing a comment and tagging someone will not tag them
- unliking/deleting a reply will not remove the notification
- no loader styling

should be notified when
behavior:
- someone likes my comment
- mentions me in a comment
- replies to my comment
- likes my shared solution
- comments on my shared solution

closes https://github.com/typehero/typehero/issues/1478
closes https://github.com/typehero/typehero/issues/1477


apps/web/src/components/mention/actions.ts
```diff
@@ -0,0 +1,17 @@
+'use server';
+
+import { prisma } from '@repo/db';
+
+export async function searchUsers(query: string) {
+  return prisma.user.findMany({
+    where: {
+      name: {
+        contains: query,
```
Hacksore, line 9: limit this to the max username length to prevent abuse.

bautistaaa: FIIIIINE

Hacksore:
looking good!

image

Only thing that is a bit annoying is github doesnt use md so you can see if I just do `@bautistaaa` (@bautistaaa) i don't need to actually interact with the autocomplete menu.

See the middle in the image above one where I just typed it 😂.

Probably it is what it is but how hard it it be to just make @ mentions work without markdown?

bautistaaa:
> looking good!
> 
> image
> Only thing that is a bit annoying is github doesnt use md so you can see if I just do `@bautistaaa` (@bautistaaa) i don't need to actually interact with the autocomplete menu.
> 
> See the middle in the image above one where I just typed it 😂.
> 
> Probably it is what it is but how hard it it be to just make @ mentions work without markdown?

I think I'd need some custom plugin to pick up `@` and convert to an anchor tag. I'll check

Hacksore:
@bautistaaa yeah you can write a custom rehype/remark plugin to do this. 

Here we have one here as an example:
https://github.com/typehero/typehero/blob/ef3a5c1242b96374c15d6284cb82577fb3c14b0b/packages/ui/src/components/markdown.tsx#L23

Hacksore:
when can i test this @bautistaaa 😂

Hacksore:
feedback so far

Fix the overflow
https://typehero-mul19j2op-trash-company.vercel.app/challenge/generic-function-arguments/comments/8?replyId=11

Fix the caching issue

Hacksore:
perhaps limit to 99+?

image

Hacksore:
minor inconsistenty is the title does not have a suffix like other pages do.
image

ex:
image

Hacksore:
suggesting to have some "not found" state rather then just super dead.

image

Hacksore:
i was able to noti myself on `SeanAWS` with this payload
```
testing @seanaws
```

do we account for case?

PR #1518, opened 2024-08-29
Fixes 1516 - Deep linking to solution comments

### LINKED ISSUE #1516: [Bug]: Linking to comments when there are multiple pages is broken on a solution
### What happened?

https://typehero.dev/challenge/generic-function-arguments/solutions/89/comments/68

you have to go to the next page to see it highlighted

appears to work fine on challenge comments. 
https://typehero.dev/challenge/generic-function-arguments/comments/59

the inifinite query is supposed to load all comments up to the reply

_No response_

### Relevant log output

_No response_

swalker326:
## Description
The sort order of the solutions comments was not set so the page in the meta data was wrong

## Related Issue
Closes #1516 

## Motivation and Context
So people can link to solution comments (only solutions were impacted)

## How Has This Been Tested?
Generated URL from sharing now correctly links and navigates to the correct page

image


bautistaaa (approved):
thanks man!

PR #1537, opened 2024-08-31
Enhancment/switch to biome

swalker326:
## Related Issue

Closes #


Hacksore:
@swalker326 when you git time let's add a CI step so we know if this is working 😂

PR #1545, opened 2024-09-03
Move og utils to a shared package

Hacksore:
Share the og utils to limit concurrent vercel builds


packages/og-utils/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@repo/og-utils",
+  "version": "0.0.0",
+  "private": true,
+  "exports": {
+    ".": "./src/index.ts"
+  },
+
+  "typesVersions": {
```
Hacksore, line 9: this makes importing work but IIRC there is a better way to do this now.

cc @anthonyshew

bautistaaa (approved):

packages/og-utils/package.json
anthonyshew: You shouldn't need `typesVersions` at all. 🤔

Hacksore: removed it in https://github.com/typehero/typehero/pull/1545/commits/35b50b4d01830603f006a402ae0863b72acdba6a lets see 🙏

Hacksore: @anthonyshew i removed it 💀
image

Hacksore: we can revisit this but i reverted it so we can ship this and figure out later.

anthonyshew: Yeah, revisit. Should be able to do that just fine...

anthonyshew (approved):

PR #1551, opened 2024-09-05
change: load monaco from cdn

bautistaaa:
Looked through the source code of the [loader](https://github.com/suren-atoyan/monaco-loader?tab=readme-ov-file#init) and compared it to some other playgrounds. It appears you can just link the `loader.js` from the CDN directly.

Prior, we were running a script that would download all of monaco and TS and place them in our `public` folder. It appears that was unnecessary. 

Hoping I can dynamically change the TS version in the playground and reload the playground accordingly


bautistaaa:
going to yolo test this in prod real quick

PR #1563, opened 2024-09-10
chore: update prisma

Hacksore:


packages/db/prisma/schema/user.prisma
```diff
@@ -0,0 +1,40 @@
+model User {
```
Hacksore, line 1: # I THINK IT IS WORKING @bautistaaa 😂

Hacksore: nvm 💀💀
image

PR #1577, opened 2024-10-07
fix: next turbo

typesafeui:
fix NextJS to use turbopack in the dev environment


bautistaaa (approved):
wtf hahaha

typesafeui:
It was hard to implement but I got it working 🙏

PR #1578, opened 2024-10-10
Fixed Enrol button bug and improved Error message formate #1571

### LINKED ISSUE #1571: [Bug]: Better error message on tracks page
### What happened?

When not logged in and you visit a tracks page to enroll you get a generic error. It should at least tell you to log in to enroll or better maybe prompt them/redirect them to login and then do the action?

Showing a proper error message is fine for now though

_No response_

### Relevant log output

```zsh
visit https://typehero.dev/tracks/typescript-foundations while not logged in.

click enroll
```

17AnuragMishra:
We have fixed the enrol button bug typehero #1571 . Users who have not signed in will be prompted to do so, and there will be no error message for such cases.

## Description
This PR fixes the issue where an error message was shown every time a user tried to enrol in a track while not logged in. Instead of just showing an error, the user can log in with a proper redirect.

I have added a check for the statusCode and message in the error handler.
A clear login prompt was provided in the toast notification.

## Related Issue
Closes #1571 

## Motivation and Context
This change is required to fix the issue where users trying to enrol in a track while not logged in would always see a generic error message. Instead of a clear prompt asking the user to log in, the system was incorrectly showing a failure to enrol message.

The fix ensures that if the user is not logged in, they are prompted to log in via a toast notification, or redirected to the login page, solving the confusion and improving user experience.

Fixes https://github.com/typehero/typehero/issues/1571


apps/web/src/app/tracks/_components/enroll-button.tsx
```diff
@@ -29,11 +34,14 @@ export function ActionButton({ action, text, trackId, slug }: EnrollButtonProps)
...
           });
-          setIsLoading(false);
-        } catch (error) {
-          error instanceof Error && error.message === 'User is not logged in'
-            ? toast({
-                title: `Please Login to continue`,
+        } catch (error: unknown) {
+          if (error instanceof Error) {
+            const customError = error as CustomError;
+            if (customError.message === 'User is not logged in' || customError.statusCode === 401) {
```
bautistaaa, line 40: its a server action so 401 probably doesnt make sense here

apps/web/src/app/tracks/_components/enroll-button.tsx
```diff
@@ -47,10 +55,32 @@ export function ActionButton({ action, text, trackId, slug }: EnrollButtonProps)
             : toast({
                 title: `Error`,
                 variant: 'destructive',
+                description: (
+                  <p>
+                    You must be logged in to enroll in a track.{' '}
+                    <a href={`/login?redirect_to=${currentUrl}`} className="font-bold underline">
```
bautistaaa, line 61: this isnt the right search param

bautistaaa:
i went ahead and fixed it

bautistaaa (approved):

PR #1585, opened 2024-10-16
perf: only query notification count if logged in

Hacksore:
Should make less queries for notification counts


typesafeui (approved):
LGTM

docker-compose.yaml
Hacksore, line 1: no longer needed

https://github.com/compose-spec/compose-spec/blob/main/spec.md#version-and-name-top-level-elements

Hacksore:
This should drop the queries we see here as well
image

PR #1586, opened 2024-10-16
fix: move logic into function for notific count

Hacksore:
Refactor to move the query guard into the db fetch method

related #1585


Hacksore:
We have a new bug now though that I thought @bautistaaa fixed where we can't tag proper cased usernames.

`@Hacksore` seems to be untaggable but `@hacksore` is

bautistaaa (approved):

bautistaaa (approved):

PR #1598, opened 2024-10-20
feat: comment actions UI improvement

### LINKED ISSUE #1597: [Feature]: Comment actions UI improvement
### Description

Currently looks cluttered, no need to show actions for every comment unless it's group-hovered with group being the entire comment block.
image

Also, too much spacing with current placement between comment and it's actions, should be less spacing than between actions and the next comment so that it's clear that actions belong to that comment and not the next one (most of that spacing is dictated by markdown `p` and `h1` tags having bottom margins and paddings, so elements below main content might need a negative margin-top, or we could move actions to the right side of the content (with 1 row like/share for other's comments and 2 row delete/edit like/share for own comments))
Screenshot 2024-10-20 at 1 30 29 AM
Screenshot 2024-10-20 at 1 31 19 AM

### Explanation

_No response_

nafisreza:
## Description
fixes #1597 

## Changes
1. Comment actions have been moved to the same line as the comment text.
2. Reply count has been moved next to edit history.
3. Actions split into two lines for better UI.
4. Like and share are always visible, the rest has a delayed transition on each button.
5. Extra bottom margin under comment text removed.

https://github.com/user-attachments/assets/82431e17-7531-48e2-ae41-bea0251dfc0c


bautistaaa:
i think we should remove the staggered animations. looked really weird when i was demoing it and my mouse was moving between lines fast. see gif

CleanShot 2024-10-21 at 16 23 48

nafisreza:
@bautistaaa then I am keeping all actions visible by default?

bautistaaa:
> @bautistaaa then I am keeping all actions visible by default?

just remove the staggered effect. basically what you had originally

nafisreza:
Here you go. @bautistaaa 

https://github.com/user-attachments/assets/023e6956-fc18-4e37-adde-e1583c42c932

bautistaaa:
awesome ill merge this soon

bautistaaa (approved):

nafisreza:
Yay, really happy to contribute. Thanks a ton guys 🤗

PR #1608, opened 2024-10-28
chore: remove aot from typehero

bautistaaa:
- redirects to new site if you try to view an aot slug or visit old aot route
  - will remove eventually
- update aot link in nav


Hacksore:
wait do you not wanna 301 to the new site?

bautistaaa:
> wait do you not wanna 301 to the new site?

ah shit you right.

Issue #1483, opened 2024-08-02
[Feature]: AOT Leaderboard


bautistaaa:
### Description

We should start implementing a leaderboard / ranking system

Easiest way is just time to submit solution converted to some point system

PR #1609, opened 2024-10-29
Feat aot overall leaderboard cache

odicho:
## Description

Implements an automatic caching strategy for the overall leaderboard using Redis:
- Cache set once 100 unique successful submissions are reached for current day
- Cache automatically invalidates at midnight EST (next challenge release)
- Reduces database load for the majority of the day

This optimization significantly improves response times for the majority of the day while ensuring accuracy during the initial submission rush.

Technical details:
- Using Redis SET with PXAT for timed expiration ([docs](https://redis.io/docs/latest/commands/set/))
- Using Redis GET ([docs](https://redis.io/docs/latest/commands/get/))

#1483


apps/aot/src/app/leaderboard/_components/daily-leaderboard.tsx
```diff
@@ -3,6 +3,8 @@ import { DataTableLeaderboard } from '@repo/ui/components/data-table-leaderboard
 import { ADVENT_CHALLENGE_IDS, LEADERBOARD_RANKING_LIMIT } from '~/app/leaderboard/constants';
 import { dailyLeaderboardColumns } from './columns';
 
+export const dynamic = 'force-dynamic';
+
 const getFirst100SubmissionsRanked = async (adventDay: number) => {
   const challengeId = ADVENT_CHALLENGE_IDS[adventDay - 1];
```
bautistaaa, line 9: I dont think we can make this a constant, id's will vary potentially, we'll need to populate this based on a db call. 

for example you can take the year from the url and find the 2024 advent of typescript track and grab all the challenge ids from there

apps/aot/src/app/leaderboard/_components/overall-leaderboard.tsx
```diff
@@ -2,16 +2,24 @@ import { Prisma, prisma } from '@repo/db';
...
 async function getOverallLeaderboard(currentAdventDay: number) {
-  const challengeIdsSoFar = ADVENT_CHALLENGE_IDS.slice(0, currentAdventDay);
+  const cachedRanking = await redisClient.get('aot-overall-leaderboard');
 
-  const ranking = await prisma.$queryRaw<OverallLeaderboardEntry[]>`
+  if (cachedRanking) {
+    return JSON.parse(cachedRanking) as OverallLeaderboardEntry[];
+  }
+
+  const challengeIdsSoFar = ADVENT_CHALLENGE_IDS.slice(0, currentAdventDay);
```
bautistaaa, line 17: this will need to be from a query

apps/aot/src/app/leaderboard/_components/overall-leaderboard.tsx
```diff
@@ -2,16 +2,24 @@ import { Prisma, prisma } from '@repo/db';
...
 import { overallLeaderboardColumns, type OverallLeaderboardEntry } from './columns';
-import { getCurrentAdventDay } from '~/utils/time-utils';
+import { getCurrentAdventDay, getNextAdventDay } from '~/utils/time-utils';
+import { redisClient } from '@repo/redis';
+
+export const dynamic = 'force-dynamic';
 
 async function getOverallLeaderboard(currentAdventDay: number) {
-  const challengeIdsSoFar = ADVENT_CHALLENGE_IDS.slice(0, currentAdventDay);
+  const cachedRanking = await redisClient.get('aot-overall-leaderboard');
```
bautistaaa, line 11: is this just going to show at the end of the event.

how would this work if we need to update the overall leaderboard day by day. the cache key is static or are we invalidating it somewhere? (Im not done reading the PR)

bautistaaa: oh nm i see it invalidates at midnight via PXAT

apps/aot/src/app/leaderboard/_components/daily-leaderboard.tsx
odicho: Hmm, are you sure? Isn't the list of challenges for AoT constant? Are you saying they're going to change?

For the queries I have below, we need challengeId's, since that's what we have an index on, not slug.

So I was thinking if we know what the challengeId's are for all of AoT, we can just have it like this:

`const ADVENT_CHALLENGE_IDS = [day_1_id, day_2_id, ...]`

and then use this as a reference. Let me know if I'm completely misunderstanding here.

apps/aot/src/utils/time-utils.ts
```diff
@@ -22,3 +22,15 @@ export function isValidAdventDay(selectedDay: number) {
   const currentAdventDay = getCurrentAdventDay();
   return selectedDay <= currentAdventDay;
 }
+
+export const getNextAdventDay = () => {
+  const currentDate = new Date();
+  const releaseTime = new Date(currentDate);
+  releaseTime.setUTCHours(5, 0, 0, 0);
+
+  if (releaseTime.getTime() <= currentDate.getTime()) {
```
bautistaaa, line 31: can you put this in english. getting confused haha

apps/aot/src/app/leaderboard/_components/overall-leaderboard.tsx
odicho: > is this just going to show at the end of the event.

Not sure I understand, but to clarify, this leaderboard will always be up-to-date.

Essentially:
1. If we have already gotten 100 unique successful submissions today, it means we have all the data we need to cache the leaderboard.
2. Cache it with expiration at next release date (see code a bit below where I do `redisClient.set`)
3. Once release days hit, we won't be able to hit cache anymore and we will query DB until first 100 unique successful submissions, and then cache again.

odicho: > how would this work if we need to update the overall leaderboard day by day. the cache key is static or are we invalidating it somewhere? (Im not done reading the PR)

`SET` has a `PXAT` option that will automatically expire the cache key at an exact timestamp.

See docs here: https://redis.io/docs/latest/commands/set/

See diff here:

apps/aot/src/app/leaderboard/_components/overall-leaderboard.tsx
```diff
@@ -33,15 +41,37 @@ async function getOverallLeaderboard(currentAdventDay: number) {
...
+    numberOfSubmissionsTodayPromise,
+  ]);
+
+  // Once we have top 100 for today, we can cache until midnight (next challenge release)
+  if (Number(numberOfSubmissionsToday) >= LEADERBOARD_RANKING_LIMIT) {
+    await redisClient.set(
+      'aot-overall-leaderboard',
+      JSON.stringify(ranking, (_, value) => (typeof value === 'bigint' ? value.toString() : value)),
+      { PXAT: getNextAdventDay() },
+    );
```
odicho, line 65: `getNextAdventDay()` returns .getTime() of next release date and time.

`PXAT` will automatically expire the cache key we set here at that time.

apps/aot/src/utils/time-utils.ts
bautistaaa: i ran this at 3:13 PT and it gave me back 10pm PT. should be be 9pm PT? Since we release at 12am ET

CleanShot 2024-10-29 at 15 14 58

odicho: It's 4am rn because of daylight time, but in December 1 it's back to EST, so it should be 5AM. Yes, I can provide some documentation here to make code more clear.

It's basically just getting the next release date/time.

odicho: So if yo want to test this in console for October, change to 4AM EDT:

```  
const releaseTime = new Date(currentDate);
 releaseTime.setUTCHours(4, 0, 0, 0);
```

odicho: I can add this as doc:

```
This function calculates when the next 5 AM UTC release will be (12:00 AM EST) - either today's 5 AM if it hasn't happened yet, or tomorrow's 5 AM if today's has already passed. 
```

odicho: To explain the `if (releaseTime.getTime() <= currentDate.getTime()) {` a bit.

1. Set releaseTime to current date, then set hour to be 5AM.
2. If we've already passed 5AM today, meaning we're at [5AM UTC - 23:59PM UTC], then we have to set release date to tomorrow's 5AM instead.
3. If we haven't passed 5AM today, meaning we're at [12:00AM UTC - 04:59AM UTC], then we we just keep it as 5AM of today since we haven't passed it yet.
4. Return .getTime()

Does that help explain?

bautistaaa (approved):
lgtm

PR #1610, opened 2024-10-29
feat: add back button on challenge page

### LINKED ISSUE #1335: [Feature]: Adding a back button on challenge/coding page
### Description

Need to add back button on the page that goes back to the challenges list page

### Explanation

Although I now know the browser back button does it, before I was always clicking the 'Explore' navbar again & again to go back to challenges list page, I think this is can be good for the UX. The behavior can be the same as the back button on the browser, I think.

After submit it can show suggestions and explore button, but I think we need this incase the user wants to go back to challenges list after seeing the detail of the challenge.

I know I'll use this feature when the challenge is damn hard for me and looking for something more easier 😂

Screenshot 2023-12-17 at 06 30 46

ansh-saini:
## Description
Added a back button on the challenge page

## Related Issue
Closes #1335 

## Motivation and Context
Fixes #1335 

## How Has This Been Tested?
Tested my changes on Google Chrome latest version. Should not need rigorous browser testing since it's a minor change.

image

### DIFF
--- a/apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
+++ b/apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
@@ -4,16 +4,18 @@ import { useMemo } from 'react';
 import { ExploreDrawer } from '~/app/challenge/_components/explore-drawer';
 import { Tooltip, TooltipContent, TooltipTrigger } from '@repo/ui/components/tooltip';
 import Link from 'next/link';
-import { ChevronLeft, ChevronRight, Swords } from '@repo/ui/icons';
+import { ArrowLeft, ChevronLeft, ChevronRight, Swords } from '@repo/ui/icons';
 import { cn } from '@repo/ui/cn';
 import { Button } from '@repo/ui/components/button';
 import { useChallengeRouteData } from '~/app/challenge/[slug]/challenge-route-data.hook';
+import { useRouter } from 'next/navigation';
 
 interface ProblemExplorerTrackNav {
   isCollapsed: boolean;
   className?: string;
 }
 export function ProblemExplorerTrackNav({ isCollapsed, className }: ProblemExplorerTrackNav) {
+  const router = useRouter();
   const { currentChallenge } = useChallengeRouteData();
   const { getTrack, title } = useProblemExplorerContext();
   const track = getTrack;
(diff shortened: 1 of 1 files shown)


apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
```diff
@@ -50,11 +52,23 @@ export function ProblemExplorerTrackNav({ isCollapsed, className }: ProblemExplo
         className,
       )}
     >
+      <Button
+        variant="ghost"
+        size="sm"
+        onClick={() => router.back()}
```
odicho, line 58: I think using .back() might cause some issues in certain scenarios:

e.g. 
- if I enter the challenge directly from a URL, .back() won't do anything and the button is non-functional.
- if I enter the challenge from any other page than explorer page, it will take me back there (do we want that?)

I think we can solve this in two different ways:
- Always show back button and always redirect to `/explore`
- Only show back button if you can do .back(), hide otherwise

cc: @bautistaaa / @Hacksore thoughts?

apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
```diff
@@ -50,11 +52,23 @@ export function ProblemExplorerTrackNav({ isCollapsed, className }: ProblemExplo
         className,
       )}
     >
+      <Button
```
odicho, line 55: I think we need to put in a bit more thought on the design here for when the panel is collapsed.

Right now, the back button is just pushing the existing button to Explore drawer to the side, which is not ideal.



Before:

https://github.com/user-attachments/assets/02d21ec6-f54b-420a-a8ec-19b2a2e8f901


After:
https://github.com/user-attachments/assets/fb198819-ab24-4425-9634-b3d32d650d43

ansh-saini: How about something like this?
image

apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
ansh-saini: 1. I think a back button should naturally lead the user back to where they came from. Hardcoding `/explore` doesn't feel like a necessary default.
2. We can have `/explore` as a default if the user landed directly on the challenge page (basically in cases where `router.back()` would be non-functional)

apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
odicho: That's definitely an improvement! I'll defer to the others to chime in here, also @PickleNik is really good with UI!

My only nit here is that I think it needs separation with a divider similar to the section below, but then you have two sections with just one icon. Not sure.

apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
odicho: Good point, but honestly not sure why we need this button if it's just going to behave exactly like the browsers built-in functionality?? @bautistaaa

apps/web/src/components/Navigation/problem-explorer-track-nav.tsx
Hacksore: Would it make sense to conditionally render it if you can go back to typehero?

ansh-saini:
Hey @odicho I'll continue to work on this PR, but can you please add the `hacktoberfest-accepted` label on this PR so it can count towards my hacktoberfest contributions?

PR #1618, opened 2024-11-01
feat(editor): support inline queries

bautistaaa:
a part of me was lost getting this to work.

we now support these two slash annotations `// ^?` `// =>`

based on the works of Orta and his chrome extension

CleanShot 2024-11-01 at 13 44 23@2x


bautistaaa:
there appears to be some clipping on the annotations on longer lines. 

HODL

image

orta:
yeah, monoco has some default limits on annotation length, its why there's whitespace compression in the playground version - it's [configurable](https://github.com/orta/vscode-twoslash-queries/blob/main/package.json#L52-L56) in the vscode extension, so probably an API somewhere in the Monaco codebase

bautistaaa:
> yeah, monoco has some default limits on annotation length, its why there's whitespace compression in the playground version - it's [configurable](https://github.com/orta/vscode-twoslash-queries/blob/main/package.json#L52-L56) in the vscode extension, so probably an API somewhere in the Monaco codebase

i've been searching for it but couldn't find anything. I saw a 120 character limit hardcoded in the ts playground plugin but it still truncated after I removed it 🙃🙃

bautistaaa:
i cant figure out the truncating thing but this is good enough ™

PR #1624, opened 2024-11-05
feat(monaco): add node types at runtime

bautistaaa:


packages/monaco/src/split-editor.tsx
```diff
@@ -311,6 +314,54 @@ export default function SplitEditor({
             const code = model.getValue();
             debouncedUserCodeAta(code);
 
+            async function addNodeTypesToMonaco() {
+              try {
+                // Fetch the main Node types index file
+                const baseUrl = 'https://unpkg.com/@types/node/';
+                const indexResponse = await fetch(`${baseUrl}index.d.ts`);
```
tylersayshi, line 321: this fetch seems like a good fit for fetching on the server with an RSC, but other than this lgtm 🚀

bautistaaa: ah!

how would that look?

fetching in parent RSC(ex: /challenge) and feeding types as props since monaco is a client component?

bautistaaa: its also worth noting we're using dta which fetches modules at runtime based on import statements

tylersayshi: > fetching in parent RSC(ex: /challenge) and feeding types as props since monaco is a client component?

yeah this is the gist... you could do the transformation that you're doing below (in Promise.all) in that parent RSC then pass it into this (nextjs then is responsible for running your fetch on your server, then JSON.stringify'ing your prop and parsing it in the client component)

> its also worth noting we're using dta which fetches modules at runtime based on import statements

what do you mean here? the fetch looks like it's always to the same url from what I see, so I am probably missing something.

bautistaaa: > what do you mean here? the fetch looks like it's always to the same url from what I see, so I am probably missing something.

now that i'm rereading i think i thought maybe there was some optimization that could be done here with RSC but this needs to happen on the client

tylersayshi: i don't follow why it "needs to happen on the client" but ¯\_(ツ)_/¯

tylersayshi:
wow ... is it just me or is github ruining my shruggie?

in editor:

image

<br />

in final ui: 

image

packages/monaco/src/split-editor.tsx
tylersayshi: yeah that's all i was thinking, but i don't think it really matters much at all here. these requests should be cheap.

bautistaaa:
will optimize later but good enough ™

PR #1630, opened 2024-11-07
docs: update sync data and uploadthing steps in local.md

rozhkoy:
Updated the steps with sync data, since the endpoint is specified there, which is not there, and also added that you need to use legacy secrets on "uploading".


LOCAL.md
```diff
@@ -48,7 +48,7 @@ GITHUB_SECRET=<client_secret>
 ### 4. Setup uploadthing for file uploads
 
 Go to [uploadthing.com](https://uploadthing.com/dashboard) and create a project,
-then copy in the secrets into your `.env`.
+then copy in the secrets (legacy, not SDK v7+) into your `.env`.
```
bautistaaa, line 51: we should probably follow up with an update to uploadthing

bautistaaa (approved):

PR #1631, opened 2024-11-07
fix: add delay for AvatarFallback

### LINKED ISSUE #1607: [Bug]: The fallback for user avatar flashes before the real avatar loads
### What happened?

fallback avatar should only display after we have confirmed the user is not logged in

_No response_

### Relevant log output

```zsh
To reproduce you can login and just refresh the home page and look at your avatar top right.

This can also be observed anywhere you see a user avatar since its reused
```

rozhkoy:
## Related Issue
closes #1607


apps/web/src/components/Navigation/index.tsx
```diff
@@ -167,7 +167,7 @@ async function LoginButton({
         >
           <Avatar className="h-7 w-7">
             <AvatarImage src={session.user.image ?? ''} alt="user avatar" />
-            <AvatarFallback>
+            <AvatarFallback delayMs={5000}>
```
bautistaaa, line 170: theres another place to do it in the `apps/aot` directory. can you take a look

bautistaaa:
i went ahead and consolidated avatar into one component to use across platforms

bautistaaa (approved):

rozhkoy:
> i went ahead and consolidated avatar into one component to use across platforms

nice

PR #1636, opened 2024-11-09
Migrate leaderboard to events route

odicho:
## Description
We are doing a couple of things in this PR:

- Moving all overall and daily AOT leaderboard to be under /events/[year]/ routes
- Adding validation and safety guards to the following routes :
    - /events/[year]
    - /events/[year]/[day]
    - /events/[year]/leaderboard
    - /events/[year]/leaderboard/[day]
- Making past leaderboards statically built (using generateStaticParams + static routes)

## How Has This Been Tested?
Navigating to all of the links above and testing edge cases:
- Try to put an invalid year in any of the routes
- Try to put an invalid day in any of the routes
- Try to put a year/day combination that is valid but challenge hasn't unlocked yet

https://github.com/user-attachments/assets/4814e841-0560-4a17-b45a-9779749cf8c4


odicho:
`/events/2023/` is an example of how to statically build past leaderboards. I will work on this further by saving the data as a .json file instead of querying the db for all past leaderboards for every build.

apps/aot/src/utils/time-utils.ts
```diff
@@ -1,35 +1,38 @@
...
 
 export function isValidAdventDay(selectedDay: number) {
-  if (selectedDay < 1 || selectedDay > 25) return false;
+  return selectedDay >= 1 && selectedDay <= 25;
+}
 
-  const currentAdventDay = getCurrentAdventDay();
-  return selectedDay <= currentAdventDay;
+export function isValidAdventYear(selectedYear: number) {
+  return selectedYear >= 2023 && selectedYear <= 2024;
```
bautistaaa, line 35: how would this work. would we need to keep adjust thing this manually?

bautistaaa (approved):

apps/aot/src/utils/time-utils.ts
odicho: Whatever is easiest tbh, maybe we can extract to some const variables somewhere? Or you can do something similar to where you queried the DB for the events and we can make sure it is part of that, I just wanted to avoid querying the db to do a simple validation.

bautistaaa: its good enough ™ for now

PR #1640, opened 2024-11-10
Adds functionality to reply to replies

### LINKED ISSUE #1605: [Feature]: Ability to reply to replies
### Description

We chose to keep conversation one level deep. However, you cannot reply to replies. Instead of disabling this the reply should be have differently and create a sibling comment that has their tag prefilled. I think thats how linkedin does it?

### Explanation

_No response_

aquibbaig:
## Description
As of now, you could only reply to parent comment. 
This PR adds the functionality to be able to add sibling comments to other replies. Additionally, when you reply to another reply comment, the "username" tag of the user you are replying to gets autofilled in the input for better UX.

## Related Issue

Closes #1605 

## Motivation and Context

The website was a bit restrictive and inconsistent where you could only reply to original comments -- this functionality was disabled in case of reply comments.  The PR adds functionality of adding comments to replies as "sibling comments" and also referencing the "recipient" of the reply for better UX.

## Implementation
I've exposed two things -- textarea element and a function to set the input value out of `CommentInput` component to control it from the parent component -- `Comment`.
- We use `setInputValue` to dynamically pass the username in a reply.
- We use the textarea element ref to control "focus" and "scroll".
- clicking on reply button shows the input + prefills the input with the `replyingTo` username.
- it also handles the edge case -- clicking on reply again when input is open should refill the input with the username.

## How Has This Been Tested?

I have tested this locally on http://localhost:3000/challenge/get-return-type which has over 150 comments. I've tested the following scenarios:
- Reply button should be visible on every comment ✅
- Clicking on reply button should display the reply text area ✅
- reply text area should auto fill the name of the user you are replying to ✅
- clicking on reply again when the input is open should re-fill the username ✅

https://github.com/user-attachments/assets/903e3b49-bc62-471f-8023-4b3243c8bfd3

### DIFF
--- a/apps/web/src/app/challenge/_components/comments/comment-input.tsx
+++ b/apps/web/src/app/challenge/_components/comments/comment-input.tsx
@@ -1,13 +1,20 @@
-import { Loader2 } from '@repo/ui/icons';
-import { useSession } from '@repo/auth/react';
-import { useEffect, useRef, useState, type RefObject } from 'react';
-import { Button } from '@repo/ui/components/button';
-import { Markdown } from '@repo/ui/components/markdown';
-import { useForm } from 'react-hook-form';
 import { zodResolver } from '@hookform/resolvers/zod';
+import { useSession } from '@repo/auth/react';
+import { Button } from '@repo/ui/components/button';
 import { Form, FormField, FormItem, FormMessage } from '@repo/ui/components/form';
-import { singleFieldSchema, type SingleFieldSchema } from '~/utils/zodSingleStringField';
+import { Markdown } from '@repo/ui/components/markdown';
+import { Loader2 } from '@repo/ui/icons';
+import {
+  forwardRef,
+  useEffect,
+  useImperativeHandle,
+  useRef,
+  useState,
+  type RefObject,
+} from 'react';
+import { useForm } from 'react-hook-form';
 import { MentionInput } from '~/components/mention/mention-input';
+import { singleFieldSchema, type SingleFieldSchema } from '~/utils/zodSingleStringField';
 
 interface Props {
   mode: 'create' | 'edit' | 'reply';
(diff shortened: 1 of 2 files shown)


bautistaaa:
damn bro is killin it. taking a look now

bautistaaa:
I think it should only prefill the tag if its replying to a reply and not the root comment. what do you think?

bautistaaa:
would you mind fixing this big gap since youre in here haha
CleanShot 2024-11-10 at 08 28 06@2x

aquibbaig:
> I think it should only prefill the tag if its replying to a reply and not the root comment. what do you think?

yeah, i missed that. its a fair assumption that you don't necessarily want to mention the user on a root comment.  

Additionally, I tried moving things around to make the comments look better.

https://github.com/user-attachments/assets/08d36775-5f74-4da1-86a5-3dfb6a724229

cc: @bautistaaa

apps/web/src/app/challenge/_components/comments/comment-input.tsx
```diff
@@ -17,103 +24,122 @@ interface Props {
...
-      await onSubmit(text);
-    } catch (e) {
-      console.error(e);
-    } finally {
-      form.reset();
-      setCommentMode('editor');
-    }
-  };
+    // We need to control the textarea element from parent component.
+    useImperativeHandle(ref, () => ({
```
bautistaaa, line 45: havent seen one of these in a long time!

apps/web/src/app/challenge/_components/comments/comment.tsx
```diff
@@ -89,6 +89,8 @@ export function Comment({
   const params = useSearchParams();
   const replyId = params.get('replyId');
 
+  const timeoutRef = useRef<NodeJS.Timeout>();
```
bautistaaa, line 92: isnt there a dom timeout type? nbd either way

bautistaaa (approved):
killed it!

PR #1647, opened 2024-11-13
feat(aot): Leaderboard UI

### LINKED ISSUE #1483: [Feature]: AOT Leaderboard

dca123:
## Description
Adds the leaderboard UI to AOT
- Needs some final touches, glimmer around the leaderboard. 

## Related Issue

Closes #1483 

image
Screenshot 2024-11-18 at 11 51 55 PM
Screenshot 2024-11-18 at 11 51 43 PM


apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx
```diff
@@ -1,14 +1,47 @@
...
+      },
+    },
+    select: {
+      name: true,
+      bio: true,
+      roles: true,
+    },
+  });
+  return userData;
+}
```
Comment (odicho) [inline apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx:42]: Suggestion:

```suggestion
async function getUserData(leaderboardEntries: OverallLeaderboardEntry[]) {
  const userIds = leaderboardEntries.map((entry) => entry.userId);

  const userData = await prisma.user.findMany({
    where: {
      id: { in: userIds },
    },
    select: {
      id: true,
      bio: true,
      roles: true,
    },
  });

  const userMap = new Map(userData.map((user) => [user.id, user]));

  return userMap;
}
```


Now you can directly user this map as a reference only when you need to render this user's bio/roles etc.

apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx
```diff
@@ -1,14 +1,47 @@
...
+      throw new Error(`Leaderboard entry for ${r.name} does not have a matching user record`);
+    }
+    return {
+      ...user,
+      ...r,
+      score: Number(r.score),
+    };
+  });
+  return rows;
+}
```
Comment (odicho) [inline apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx:27]: In tandem with the suggestion below, I think this can be removed, and then you can use the leaderboard data to build the table, and for each row just directly reference the map we built below. Every user will have a guaranteed entry in the map.

apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx
```diff
@@ -23,7 +56,7 @@ async function getOverallLeaderboard(year: number, isPast: boolean) {
   SELECT
     u.name,
```
Comment (odicho) [inline apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx:57]:
```suggestion
  SELECT
    u.id as userId
    u.name,
```

This will help with user data lookup, using the primary key.

Comment (odicho) [reply apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx:57]: Also make sure to update: `OverallLeaderboardEntry` to include `userId`!

apps/aot/src/app/events/[year]/leaderboard/_components/daily-leaderboard.tsx
```diff
@@ -24,21 +27,21 @@ const getFirst100SubmissionsRanked = async (adventYear: string, adventDay: strin
...
-    </div>
+const calculateDuration = (start: { year: string; day: string }, end: Date) => {
+  const duration = differenceInMilliseconds(
+    end,
+    new Date(Number(start.year), 11, Number(start.day)),
   );
-}
+  const date = format(duration, 'HH:MM:SS');
+  return date;
+};
```
Comment (odicho) [inline apps/aot/src/app/events/[year]/leaderboard/_components/daily-leaderboard.tsx:47]: Not sure I fully understand the purpose of this function! Are you able to elaborate?

apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx
Comment (dca123) [reply apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx:42]: Is there a guarantee that the order of user data will be ===. order of userids ?

apps/aot/src/app/events/[year]/leaderboard/_components/daily-leaderboard.tsx
Comment (dca123) [reply apps/aot/src/app/events/[year]/leaderboard/_components/daily-leaderboard.tsx:47]: This is to show the time taken to solve the challenge. The database has '2024-01-25T06:07:44.533Z' as created at, and I want to show how many HH:MM:SS has it elapsed since the start of the challenge day, to get to this time.

apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx
Comment (odicho) [reply apps/aot/src/app/events/[year]/leaderboard/_components/overall-leaderboard.tsx:43]: It doesn't have to be in the same order! When you iterate through the leaderboard entries, you can lookup the info from the map!

apps/aot/src/app/events/[year]/leaderboard/_components/daily-leaderboard.tsx
Comment (odicho) [reply apps/aot/src/app/events/[year]/leaderboard/_components/daily-leaderboard.tsx]: Feel free to use `formatDate` function here: 

https://github.com/typehero/typehero/blob/94fe390088b79949ffa089f35387486fb0dd06e0/apps/aot/src/app/events/%5Byear%5D/leaderboard/_components/columns.tsx#L74

bautistaaa:
kinda weird to waste all the space on the left i think?

image

apps/aot/src/app/events/2023/leaderboard/_components/table.tsx
```diff
@@ -0,0 +1,66 @@
...
+            <TableCell className="text-center ">{i + 4}</TableCell>
+            <TableCell className="flex flex-col items-center justify-center gap-1 text-center md:flex-row md:gap-2">
+              <UserBadge
+                className="text-wrap break-all underline-offset-4 group-hover:underline"
+                user={{
+                  ...d,
+                  image: d.image ?? '',
+                }}
+              />
+              {d.roles.find((r) => r.role === RoleTypes.SUPPORTER) !== undefined ? (
```
bautistaaa, line 48: `.some` will return a boolean for ya

apps/aot/src/app/events/[year]/leaderboard/[day]/page.tsx
```diff
@@ -1,7 +1,5 @@
-import DailyLeaderboard from '../_components/daily-leaderboard';
-
 // export const revalidate = 10; Need next.js v15
 
 export default function Page({ params: { year, day } }: { params: { year: string; day: string } }) {
-  return <DailyLeaderboard adventYear={year} adventDay={day} />;
+  return <></>;
```
Comment (bautistaaa) [inline apps/aot/src/app/events/[year]/leaderboard/[day]/page.tsx:4]: if theres nothing in here I think we can delete it or does it break?

my nextjs fu is bad

bautistaaa:
i think we have a error.tsx we can use if this blows up for some reason. unsure if it will traverse and find the root one or if we need to define it in the directory

apps/aot/src/app/events/[year]/leaderboard/[day]/page.tsx
Comment (dca123) [reply apps/aot/src/app/events/[year]/leaderboard/[day]/page.tsx:4]: nah the leaderboard is only implemented for 2023 rn, lemme add it to this page too.

bautistaaa:
i think we can probably merge this ASAP and make incremental improvements

bautistaaa (approved):

PR #1648, opened 2024-11-13
feat(comments): port comment enhancements from web

bautistaaa:
copy pasta from web until i consolidate the madness

still need to test


bautistaaa:
fixed the shareable link issue in later commit.

it can be shipped now 🚀

PR #1649, opened 2024-11-14
fix(body): hide the horizontal scrollbar

malkiii:
## Description
There's some overflowing issue that displays the scrollbar.

Screenshot 2024-11-14 093021

### DIFF
--- a/apps/web/src/styles/globals.css
+++ b/apps/web/src/styles/globals.css
@@ -111,7 +111,7 @@
     @apply border-border;
   }
   body {
-    @apply bg-background text-foreground;
+    @apply bg-background text-foreground overflow-x-hidden;
   }
 }
 


bautistaaa:
I think the real fix is to figure out why the hero is causing the overflow. removing the hero removes the scroll bar

malkiii:
> I think the real fix is to figure out why the hero is causing the overflow. removing the hero removes the scroll bar

yeah I figured it out, it's this right container that causes the overflow: 
https://github.com/typehero/typehero/blob/64e7b99f6b15af0198acc45bffcd0af3e7c98840/apps/web/src/app/_components/hero-illustration.tsx#L29
the problem is that you have to remove all the elements inside to hide the scrollbar, and if you try to add `overflow-x-hidden` to one of the containers above it will display a vertical scrollbar, so I think the only solution for this is:
- add `overflow-hidden` to the `<section>` element.
- add `overflow-x-hidden` to the body.
- or just try to change the styles of the elements inside.

bautistaaa:
going to close for now but feel free to re open if you find a way forward

PR #1657, opened 2024-11-22
quick fix to aot badges for 2024

bautistaaa:
CleanShot 2024-11-21 at 16 00 51@2x


bautistaaa:
long term we'll refactor all this to work in both platforms and can probably do it all in sql

PR #1662, opened 2024-11-23
fix day numbers and also make query to get challenges much more succinct

odicho:
fixed the query so that we can get the day, previously using the id would not equal to the actual day of advent.

We can also use select in the query and it will only get what is necessary for us.

before:

```
{
aot:dev:     id: <redacted>,
aot:dev:     createdAt: 2024-10-26T20:23:40.985Z,
aot:dev:     updatedAt: 2024-11-20T01:10:41.166Z,
aot:dev:     difficulty: 'EVENT',
aot:dev:     name: '2024 Day Nine',
aot:dev:     slug: '2024-9',
aot:dev:     description: <redacted>,
aot:dev:     shortDescription: '2024 Day Nine of Advent of Typescript',
aot:dev:     code: <redacted>,
aot:dev:     tests: <redacted>,
aot:dev:     status: 'ACTIVE',
aot:dev:     tsconfig: { strict: true },
aot:dev:     userId: <redacted>,
aot:dev:     user: { name: 'TypeHero' },
aot:dev:     hasSolved: false,
aot:dev:     active: true
aot:dev:   }
```

after:

```
{ day: 25, hasSolved: false, active: true }
```

### DIFF
--- a/apps/aot/src/server/api/routers/event.ts
+++ b/apps/aot/src/server/api/routers/event.ts
@@ -98,9 +98,10 @@ export const eventRouter = createTRPCRouter({
                 lte: currentAdventDay - 1,
               },
             },
-            include: {
+            select: {
+              orderId: true,
               challenge: {
-                include: {
+                select: {
                   submission: {
                     where: {
                       userId: ctx.session?.user?.id || '',
@@ -110,11 +111,6 @@ export const eventRouter = createTRPCRouter({
                       isSuccessful: true,
                     },
                     take: 1,
-                  },
-                  user: {
-                    select: {
-                      name: true,
-                    },
                   },
                 },
               },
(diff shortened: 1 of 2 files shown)


odicho:
> do we need to spread challenge in the return since we're only using calculated/derived values

good callout, let me update that as well!

apps/aot/src/server/api/routers/event.ts
```diff
@@ -98,38 +98,28 @@ export const eventRouter = createTRPCRouter({
...
-          ...challenge,
-          hasSolved: trackChallenge.challenge.submission.length > 0,
-          active: true,
-        };
-      });
+      return challenges?.trackChallenges.map((trackChallenge) => ({
+        day: trackChallenge.orderId + 1,
+        hasSolved: trackChallenge.challenge.submission.length > 0,
+        active: true,
+      }));
```
odicho, line 123: @bautistaaa let me know if this looks good to you!

bautistaaa (approved):

PR #1673, opened 2024-11-28
feat: Add Year switcher + socials to nav bar

dca123:
## Description
Moves the year switcher to the navigation bar. When changing years, the visible Events and Leaderboard links have their urls updated to the correct year. If the user is on an event page, i.e /events/2024/leaderboard, switching the year to 2023, will send them to the same page on 2024.

The live ping on 2024 will appear while the event is running. 

## How Has This Been Tested?
Checked it on both mobile + desktop

image
image
image


apps/aot/src/components/landing/Partners.tsx
```diff
@@ -38,7 +38,7 @@ export default function Partners() {
         </div>
         <div className="infinite-scroll-x flex flex-nowrap items-center md:hidden">
           {[...partners, ...partners].map((partner, i) => {
-            return <PartnerLink key={`infinite-partners-${partner.name}-${i}`} partner={partner} />;
+            return <PartnerLink key={`${partner.name}_${i}`} partner={partner} />;
```
dca123, line 41: fixes key issue

bautistaaa (approved):

PR #1675, opened 2024-11-30
fix(aot): make text white on light mode

Pyr33x:
This is for better visibility 

### Before
Screenshot 2024-11-30 at 3 02 16 PM

### After
Screenshot 2024-11-30 at 3 02 39 PM


bautistaaa:
already fixed/changed!

PR #1679, opened 2024-11-30
fix: initial quick accessibility fixes

### LINKED ISSUE #1678: [Bug]: Assorted accessibility issues on AoT 2024
### What happened?

1. Run Lighthouse's Accessibility check (i.e. aXe)
2. See complaints:
    1. [Buttons do not have an accessible name](https://dequeuniversity.com/rules/axe/4.10/button-name)
    2. [Links do not have a discernible name](https://dequeuniversity.com/rules/axe/4.10/link-name)

Additionally, I can tell from using the site that the animations don't respect [prefer-reduced-motion](https://web.dev/articles/prefers-reduced-motion).

All

### Relevant ~~log output~~ info

When buttons or links don't have an accessible name, folks who can't visually see the button won't have any way of knowing what they are.

Reduced motion is important to avoid distracting folks with ADHD or other conditions. Also for folks on reduced hardware / with limited batteries.

JoshuaKGoldberg:
## Description

Applies a few initial accessibility fixes.

## Related Issue

Closes #1678

## Motivation and Context

Because I want the AoT site to be usable by everyone, not just folks with full, uninhibited brain+motor+vision control. 🙂 

You can read deeper explanations here: https://web.dev/accessibility.

## How Has This Been Tested?

1. Running Lighthouse's Accessibility check (i.e. aXe)
2. <kbd>Tab</kbd>bing through the site with a screenreader

See inline comments.


apps/aot/src/components/navigation/index.tsx
JoshuaKGoldberg, line 1: For color contrast, it was that the `text-black/50` in the header wasn't contrast-y enough against the header (4:1 instead of 4.5:1 contrast ratio):

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Before capture of 50% black on white for the 2024 year dropdown
</td>
<td>
After capture of 60% (more visible) black on white for the 2024 year dropdown
</td>
</tr>
</tbody>
</table>

bautistaaa (approved):
appreciate it!

apps/aot/src/components/navigation/index.tsx
```diff
@@ -35,27 +35,30 @@ export async function Navigation() {
...
             href="/support"
           >
             Support Us
           </Link>
           <Separator className="md:hidden dark:bg-white/20" />
           <NavLink
             href={DISCORD_INVITE_LINK}
-            className="flex flex-row items-center gap-2 text-black/40 md:hidden dark:text-white/60"
+            className="flex flex-row items-center gap-2 text-black/40 dark:text-white/60"
+            linkClassName="md:hidden"
```
JoshuaKGoldberg, line 47: `NavLink` passes `className` to a child rendered inside a `<Link>`. That means the `<Link>` is not hidden when the child's `md:hidden` is in effect. A `<Link>` with no children still gets <kbd>Tab</kbd> focus, which is an accessibility no-no.

Aside: this is one of the reasons why having a `className` prop apply to a _child_ component, rather than the root of what's rendered, is confusing. I recommend against it. 😬

apps/aot/src/components/CountdownTimer.tsx
```diff
@@ -21,6 +21,10 @@ const useAotCountdown = () => {
     }
 
     const countdown = () => {
+      if (window.matchMedia('(prefers-reduced-motion: reduce)').matches) {
+        return;
+      }
```
JoshuaKGoldberg, line 26: For this and the other `prefers-reduced-motion` changes: https://web.dev/articles/prefers-reduced-motion.

It would be better to add a play/pause button for animations: https://www.w3.org/WAI/tutorials/carousels/animations. That way people (like me!) who have a hard time using a using a webpage when there are distracting animations don't have to opt out of them altogether.

apps/aot/src/components/navigation/index.tsx
bautistaaa: yeah thats odd. will followup and clean that up later

apps/aot/src/components/CountdownTimer.tsx
bautistaaa: ohh TIL

apps/aot/src/components/navigation/logo.tsx
```diff
@@ -1,6 +1,12 @@
 export default function Logo() {
   return (
-    <svg className="h-6" viewBox="0 0 569 123" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <svg
+      aria-label="advent<TS>"
```
JoshuaKGoldberg, line 4: When an `<svg>` or other visual asset doesn't have a label or title, there's no way  to know what it is programmatically. This is bad both for screenreaders and for SEO.

https://web.dev/learn/accessibility/images/

PR #1683, opened 2024-12-01
fix: reset About and Support buttons back to default focus

### LINKED ISSUE #1682: [Bug]: AoT About and Support page buttons don't show enough of a focus indicator
### What happened?

When an element has focus, such as from <kbd>Tab</kbd> focus, that should be visually apparent: https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html. Otherwise folks who are using a keyboard or other non-mouse device can't easily tell what they've currently got focused.

That focus indication needs to have enough color contrast that low vision folks can see it: https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html

The buttons in About and Support override the default focus styles in a way that's super subtle, and for at least the donation button, definitely not AA compliant:

https://github.com/user-attachments/assets/b034373d-87f0-4d9b-a9ad-850e9ab73906

All

### Relevant log output

```zsh
Aside: https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html is a _beautiful_ thing for showing focus indications only when the user probably wants them.
```

JoshuaKGoldberg:
## Description

Removes removals of default focus-visible outlines for the buttons.

I also fixed their widths on mobile to be consistently 100% width and centered.

## Related Issue

Closes #1682

## Motivation and Context

Adding in the default focus state makes it so the `:focus-visible` view of the buttons is color contrast compliant.

## How Has This Been Tested?

I only tested this stuff on the AoT site.

**You should not trust me, I didn't test these anywhere else!** I don't even know how or where I _would_ test them. 😁


bautistaaa (approved):
looks good on both sites!

PR #1687, opened 2024-12-01
fix: preliminary accessibility issues in editor

### LINKED ISSUE #1686: [Bug]: Assorted accessibility issues on AoT 2024 editor
### What happened?

1. Visit https://www.adventofts.com/events/2024/1
2. In a Monaco code editor, press <kbd>F1</kbd>, type something like `high contrast`, and click the result or hit <kbd>Enter</kbd> to enable Monaco's high contrast theme[^monaco-a11y]
3. Run Lighthouse's Accessibility check (i.e. aXe)
4. See complaints:
    1. [Buttons do not have an accessible name](https://dequeuniversity.com/rules/axe/4.10/button-name)
    2. [Background and foreground colors do not have sufficient color contrast](https://dequeuniversity.com/rules/axe/4.10/color-contrast)

All

### Relevant ~~log output~~ info

When text doesn't have enough color contrast against its background, folks with limited vision -such as those with bad eyesight or lots of screen glare- might not be able to read the text.

When buttons or links don't have an accessible name, folks who can't visually see the button won't have any way of knowing what they are.

[^monaco-a11y]: Monaco has a rich accessibility story, courtesy of Microsoft needing to make VS Code accessible: https://github.com/Microsoft/monaco-editor/wiki/Monaco-Editor-Accessibility-Guide.

JoshuaKGoldberg:
## Description

Applies a few initial accessibility fixes.

## Related Issue

Closes #1686.

## Motivation and Context

🙌 accessibility! 🙌

## How Has This Been Tested?

**You should not trust me on this.** I don't have the editor running locally (I hate setting up auth+databases 😛) and was applying these changes manually to the DOM. I also only verified them in the AoT editor, not the rest of TypeHero.

See inline comments.

### DIFF
--- a/apps/aot/src/app/events/(leaderboard)/_components/leaderboard-page.tsx
+++ b/apps/aot/src/app/events/(leaderboard)/_components/leaderboard-page.tsx
@@ -21,7 +21,7 @@ export function LeaderboardPage(props: {
         </div>
       </div>
       <div className="border-[hsla(0, 0%, 100%, 0.12)] relative mx-auto mt-[255px] max-w-screen-lg rounded-2xl rounded-b-none border border-b-0 bg-[hsla(0,0%,0%,0.07)] px-1 pt-1 shadow-[0px_-18px_131px_-78px_hsla(132,100%,53%,0.3)] backdrop-blur-sm md:mt-[350px] lg:mt-[500px] dark:bg-[hsla(0,0%,100%,0.07)]">
-        <div className={`${styles['ring-container']} -inset-0`}>
+        <div className={`${styles['ring-container']} pointer-events-none -inset-0`}>
           <div className={`${styles['ring-animation']} inset-1 bg-slate-50/10`} />
         </div>
         <div className="bg-background rounded-lg rounded-b-none">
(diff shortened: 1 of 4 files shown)


apps/aot/src/app/globals.css
```diff
@@ -22,7 +22,7 @@
     --border: 240 5.9% 90%;
     --input: 240 5.9% 90%;
 
-    --primary: 211 61% 48%;
+    --primary: 211 61% 47%;
```
JoshuaKGoldberg, line 25: The color contrast for this blue is _really_ close to the minimum 4.5:1. But it's not quite there:

* Before (4.35:1): https://webaim.org/resources/contrastchecker/?fcolor=FAFAFA&bcolor=3078C5
* After (4.53:1): https://webaim.org/resources/contrastchecker/?fcolor=FAFAFA&bcolor=2F75C1

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td>
Screenshot of before 'Submit' button with just barely not dark enough blue for contrast
</td>
<td>
Screenshot of after 'Submit' button with just barely dark enough blue for contrast
</td>
</tr>
</tbody>
</table>

apps/aot/src/app/events/[year]/[day]/_components/description/index.tsx
```diff
@@ -89,7 +89,7 @@ export function Description({ challenge }: Props) {
                     'rounded-full',
                   )}
                 >
-                  <Share className="h-4 w-4" />
+                  <Share aria-label="Share" className="h-4 w-4" />
```
Comment (JoshuaKGoldberg) [inline apps/aot/src/app/events/[year]/[day]/_components/description/index.tsx:92]: These elements don't provide their own label or title themselves. So there's nothing in them by default that says what they are in HTML.

bautistaaa (approved):

PR #1689, opened 2024-12-01
fix: make leaderboard links clickable

### LINKED ISSUE #1688: [Bug]: Usernames in the leaderboard aren't clickable even though they're links
### What happened?

You can't click rows in the leaderboard to view their profiles even thought they're links. I think this happens because the ring animation has `position: absolute` and is above the leaderboard element.

https://github.com/user-attachments/assets/38d966d4-a464-4123-ab78-b8aac27347b9

Arc

### Relevant log output

_No response_

soorria:
Make leaderboard links clickable

## Description
Adds `pointer-events: none` to the ring animation element so that the leaderboard that's behind it is clickable. Could probably also be fixed w/ `z-index` or `position: relative` on the leaderboard element.

Lmk if i should change what i've done!

## Related Issue
Closes #1688

## Motivation and Context
The user names seem like they should be clickable (they're links!), but you can't.

## How Has This Been Tested?
Tested by manually editing the styles in the dev tools. I couldn't install packages due to some patches not being applied :confused:.

Video from issue showing it working after adding `pointer-events: none`:

https://github.com/user-attachments/assets/6184c72b-91f6-4bcc-a9b0-9b2602216f83


bautistaaa:
omg thanks. i just reported this

bautistaaa:
can you run `pnpm checks:fix`

bautistaaa (approved):

PR #1716, opened 2024-12-26
Evaluate code at time of submission

### LINKED ISSUE #1695: [Bug]: Initially the solution is always marked incorrect
### What happened?

When the challenge has been just opened, the solution is initially marked incorrect, so if you hit submit too fast the answer will be marked as incorrect
image

Steps to reproduce:
- Write a solution while logged out
- Login
- Go to the challenge and hit submit fast

Suggested solution:
Have a loading state when Monaco is loading all the needed types, and disable the submit button in that state

_No response_

### Relevant log output

_No response_

nickbar01234:
## Description

The changes in this PR re-evaluates `tsErrors` at time of submission. This is a workaround to #1695, where the solution is rejected on first log-in. Note that this solution will solve a similar class of problem, where users make changes to the code and quickly press submit.

## Looking Forward

As a next step, I think it would be nice to capture the code at the time of submission and only run the type-check on this code. The current changes does not account for the fact that users may change the code after pressing the submit button - however, I reason that this is quite unlikely.

## Related Issue

Closes #1695

## How Has This Been Tested?

1. Ensure that the original scenario in #1695 is not reproduceable
2. Add a random import statement and quickly press submit

cc @typesafeui


nickbar01234:
Unrelated to this PR, but I'm confused by what [this snippet](https://github.com/typehero/typehero/blob/88492c2f9f119f563528ebef033ab7120b641f1d/packages/monaco/src/index.tsx#L209-L219) is doing.

```
const testErrors = await Promise.all([
  tsWorker.getSemanticDiagnostics(USER_CODE_PATH),
  tsWorker.getSyntacticDiagnostics(USER_CODE_PATH),
  tsWorker.getCompilerOptionsDiagnostics(USER_CODE_PATH),
] as const);

const userErrors = await Promise.all([
  tsWorker.getSemanticDiagnostics(USER_CODE_PATH),
  tsWorker.getSyntacticDiagnostics(USER_CODE_PATH),
  tsWorker.getCompilerOptionsDiagnostics(USER_CODE_PATH),
] as const);
```

Seems like we should be evaluating `TESTS_PATH` here?

packages/monaco/src/index.tsx
```diff
@@ -134,14 +161,17 @@ export function CodePanel(props: CodePanelProps) {
         action: <ToastAction altText="Dismiss">Dismiss</ToastAction>,
       });
     }
-  }, [tsErrors]);
+  }, [code, monacoInstance, disabled]);
+
+  const debouncedHandleSubmit = useCallback(debounce(handleSubmit, 1000), [handleSubmit]);
```
bautistaaa, line 166: can we do 500

nickbar01234: Sounds good, pushed the new changes

bautistaaa:
> Unrelated to this PR, but I'm confused by what [this snippet](https://github.com/typehero/typehero/blob/88492c2f9f119f563528ebef033ab7120b641f1d/packages/monaco/src/index.tsx#L209-L219) is doing.
> 
> ```
> const testErrors = await Promise.all([
>   tsWorker.getSemanticDiagnostics(USER_CODE_PATH),
>   tsWorker.getSyntacticDiagnostics(USER_CODE_PATH),
>   tsWorker.getCompilerOptionsDiagnostics(USER_CODE_PATH),
> ] as const);
> 
> const userErrors = await Promise.all([
>   tsWorker.getSemanticDiagnostics(USER_CODE_PATH),
>   tsWorker.getSyntacticDiagnostics(USER_CODE_PATH),
>   tsWorker.getCompilerOptionsDiagnostics(USER_CODE_PATH),
> ] as const);
> ```
> 
> Seems like we should be evaluating `TESTS_PATH` here?

thats probably a bug or this is overridden by something check test path downstream

nickbar01234:
> thats probably a bug or this is overridden by something check test path downstream

This is called when user-editor first mounts; so yeah it makes sense that the test editor mounting later overrides it.

bautistaaa (approved):
lgtm

we'll add a server layer in 2025

PR #1717, opened 2024-12-28
Enforce submission is only viewable by author

### LINKED ISSUE #1715: Submissions should only be viewable by its author
### What happened?

anyone can see a submission by plugging in random ids. it should only be viewable by the author. you have to explicitly share them to be public 

_No response_

### Relevant log output

_No response_

nickbar01234:
## Description

Query submission using current session user, which enforces that submission is only viewable by the author.

## Related Issue

Closes #1715.

## How Has This Been Tested?

* Create a submission and grab its id
* Log out and navigate to `localhost:3000/challenge/<problem>/submissions/<id>`

image

cc @bautistaaa


nickbar01234:
Did not port the changes for aot project, which I assume is not necessary since it's over. Also, not much of a frontend person, so lmk if there's a specific design I should follow with "Submission does not exist"

bautistaaa (approved):

bautistaaa:
ah shoot it should def be there, you can just duplicate what you did. i'll consolidate all of this later

PR #1722, opened 2025-01-11
style: Remove pink blur and apply green for Learners section in explore page

### LINKED ISSUE #1721: [Bug]: Learners section title background color does not match tag color
### What happened?

I think the background color of each section's title **should match the tag color**.
However, in the **Learners section, the background color appears pink** instead of green, which seems to be a bug.

Circled in red
Screenshot 2025-01-11 at 4 54 22 PM

All

### Relevant log output

_No response_

heygwangjin:
## Description
I fixed the issue where the title background color of the Learners section appeared pink instead of green on the `/explore` page route.

Please see the screenshot below.

## Related Issue

Closes #1721 

## How Has This Been Tested?

After testing on all browsers, the green background works as expected.

### Before
Screenshot 2025-01-11 at 11 12 11 PM

### After
Screenshot 2025-01-11 at 11 11 54 PM


bautistaaa (approved):
thanks, i cant even tell its there 🤣

